PR #79, opened 2019-11-11
feat(main): add application icon

ysfscream:


package.json
```diff
@@ -1,12 +1,14 @@
 {
   "name": "MQTTX",
+  "author": "emqx",
```
CrazyWisdom: EMQ X Team

ysfscream: Fixed

PR #250, opened 2020-06-24
fix(ConnectionForm.vue): add deep watch and change editor fontSize

0721Betty:


src/components/Editor.vue
```diff
@@ -69,6 +69,9 @@ export default class Editor extends Vue {
     if (!id) {
       return false
     }
+    if (this.id === 'lastWillPayload') {
+      defaultOptions.fontSize = 12
+    }
```
ysfscream, line 74: You should use `@Prop` to set `font-size`, and if this attribute is not passed in, the default value is `14`, this will make this component more scalable.

src/components/Editor.vue
```diff
@@ -14,6 +14,7 @@ import EditorNight from '@/assets/scss/theme/editor-night.json'
 export default class Editor extends Vue {
   @Prop({ required: true }) public id!: string
   @Prop({ required: true }) public lang!: string
+  @Prop({ required: true }) public fontSize!: number
```
ysfscream, line 17: @Prop({ required: true, default: 14 }) public fontSize: number

src/components/MsgPublish.vue
```diff
@@ -41,6 +41,7 @@
         ref="payloadEditor"
         id="payload"
         :lang="payloadLang"
+        :fontSize="14"
```
ysfscream, line 44: Remove

src/components/Editor.vue
ysfscream: `fontSize` has default value instead of required.

Issue #394, opened 2020-12-01
Append timestamp to MQTT client id (Optional)


ysfscream:

oceanlvr:
@ysfscream Hi! Can we close this issue now? :raised_hand::raised_hand:

ysfscream:
@oceanlvr No now, close it when the new version package containing this issue is released.

ysfscream:
Fixed on [v1.4.2](https://github.com/emqx/MQTTX/releases/tag/v1.4.2)

PR #401, opened 2020-12-10
feat(connection): support append timestamp to MQTT client id.

oceanlvr:
Feature realization:

* Support append timestamp to MQTT client ID, which prevent duplicate connections.
* Add floating tips to explain what the client ID timestamp button does.

The timestamp of the attachment will not be displayed on the connection interface, but will be attached to the clientID when actually connecting.

connect success

And two connections with the same client ID will exist at the same time due to attachment time.(Actually not the same clientID)

Two connections with the same id will exist at the same time due to attachment time

[https://github.com/emqx/MQTTX/issues/394#event-4082449360](https://github.com/emqx/MQTTX/issues/394#event-4082449360)

### DIFF
--- a/src/utils/mqttUtils.ts
+++ b/src/utils/mqttUtils.ts
@@ -52,7 +52,7 @@ export const getClientOptions = (record: ConnectionModel): IClientOptions => {
   }
   options.connectTimeout = time.convertSecondsToMs(connectTimeout)
   // Append timestamp to MQTT client id
-  if (clientIdWithTime === true) {
+  if (clientIdWithTime) {
     const clickIconTime = Date.parse(new Date().toString())
     options.clientId = `${options.clientId}_${clickIconTime}`
   }
--- a/src/utils/mqttUtils.ts
+++ b/src/utils/mqttUtils.ts
@@ -38,6 +38,7 @@ export const getClientOptions = (record: ConnectionModel): IClientOptions => {
     reconnect,
     will,
     rejectUnauthorized,
+    clientIdWithTime,
   } = record
   // reconnectPeriod = 0 disabled automatic reconnection in the client
   const reconnectPeriod = reconnect ? 4000 : 0
@@ -50,6 +51,11 @@ export const getClientOptions = (record: ConnectionModel): IClientOptions => {
     protocolVersion,
   }
   options.connectTimeout = time.convertSecondsToMs(connectTimeout)
+  // Append timestamp to MQTT client id
+  if (clientIdWithTime === true) {
+    const clickIconTime = Date.parse(new Date().toString())
+    options.clientId = `${options.clientId}_${clickIconTime}`
+  }
   // Auth
   if (username !== '') {
     options.username = username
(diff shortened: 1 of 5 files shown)


ysfscream:
I think your PR commit did not pull the upstream branch dev/1.4.2 to the latest, beacuse the file `manual-cn.md` has been modified in the previous commit. Please check.

src/views/connections/ConnectionForm.vue
```diff
@@ -56,11 +56,22 @@
...
+            <!-- add clientID timestamp check icon -->
+            <el-col :span="1">
+              <a
+                href="javascript:;"
+                class="icon-oper"
+                @click="appendClientIDWithTime"
+                :class="{ 'icon-oper-active': clientIdWithTime }"
+              >
+                <i class="el-icon-time"></i>
+              </a>
```
ysfscream, line 73: Please add a `Tooltip` component in order to explain what this button does.

src/views/connections/ConnectionForm.vue
```diff
@@ -566,8 +582,21 @@ export default class ConnectionCreate extends Vue {
...
+    this.record.clientIdWithTime = !clientIdWithTimeStatus
+    if (this.record.clientIdWithTime === true) {
+      //parse to unix datetime
+      const clickIconTime = Date.parse(new Date().toString())
+      this.record.clientId = `${this.record.clientId}_${clickIconTime}`
+    } else {
+      const curClientID = this.record.clientId
+      this.record.clientId = curClientID.substring(0, curClientID.lastIndexOf('_'))
+    }
+  }
```
ysfscream, line 599: Append method should not be implemented here. Should be added when creating MQTT client options.

ysfscream (changes requested):

src/views/connections/ConnectionForm.vue
```diff
@@ -56,11 +56,31 @@
...
-            <el-col :span="2">
+            <el-col :span="1">
               <a href="javascript:;" class="icon-oper" @click="setClientID">
                 <i class="el-icon-refresh-right"></i>
               </a>
             </el-col>
+            <!-- add clientID timestamp check icon -->
+            <el-col :span="1">
+              <el-tooltip
+                v-if="oper === 'create'"
```
ysfscream, line 67: Remove `v-if`, because it can also be modified when editing.

ysfscream (changes requested):

ysfscream:
Could you please add this feature to the `ConnectionInfo` component? You can add this time icon next to the ClientID label as shown below.

src/utils/mqttUtils.ts
```diff
@@ -50,6 +51,11 @@ export const getClientOptions = (record: ConnectionModel): IClientOptions => {
     protocolVersion,
   }
   options.connectTimeout = time.convertSecondsToMs(connectTimeout)
+  // Append timestamp to MQTT client id
+  if (clientIdWithTime === true) {
```
0721Betty, line 55: if(clientIdWithTime)?

ysfscream: Is a good suggestion.

ysfscream (approved):

0721Betty (approved):

PR #409, opened 2020-12-14
feature(connections): disable ConnectionInfo table editing when client was connected

oceanlvr:
The following is the previous version:

before

We expect that the form cannot be modified when connected, Now we cannot modify the form while connected.

after

### DIFF
--- a/src/views/connections/ConnectionInfo.vue
+++ b/src/views/connections/ConnectionInfo.vue
@@ -31,7 +31,7 @@
                 slot="suffix"
                 title="Refresh"
                 class="el-icon-refresh"
-                :class="client.connected ? 'icon-oper-disable' : ''"
+                :class="{ 'icon-oper-disable': client.connected }"
                 @click="refreshClientId"
               >
               </i>
@@ -239,6 +239,10 @@ export default class ConnectionInfo extends Vue {
       // icon disable
       .icon-oper-disable {
         color: var(--color-text-default);
+      }
+      // input disable
+      .is-disabled > input {
+        background: transparent;
       }
     }
     // adjust the position of the clientID timestamp element
--- a/src/views/connections/ConnectionInfo.vue
+++ b/src/views/connections/ConnectionInfo.vue
@@ -4,7 +4,7 @@
       <el-row :gutter="20">
         <el-col :span="8">
           <el-form-item :label="$t('connections.name')" prop="name">
-            <el-input size="mini" v-model="connection.name"></el-input>
+            <el-input size="mini" v-model="connection.name" :disabled="client.connected"></el-input>
           </el-form-item>
         </el-col>
         <el-col :span="8">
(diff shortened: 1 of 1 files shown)


src/views/connections/ConnectionInfo.vue
```diff
@@ -139,6 +151,11 @@ export default class ConnectionInfo extends Vue {
 
   // Reverse the status of clientIdWithTime.
   private reverseClientIDWithTime() {
+    // when the client connected, we should disable this icon event
+    let curClientConnection: any = this.client
```
ysfscream, line 155: `this.client` is not the `any` type.

src/views/connections/ConnectionInfo.vue
```diff
@@ -179,6 +196,11 @@ export default class ConnectionInfo extends Vue {
   }
 
   private refreshClientId() {
+    // when the client connected, we should disable this icon event
+    let curClientConnection: any = this.client
+    if (curClientConnection.connected === true) {
+      return
+    }
```
ysfscream, line 203: The code here is duplicated, please consider modifying it as a function.

src/views/connections/ConnectionInfo.vue
```diff
@@ -139,6 +151,11 @@ export default class ConnectionInfo extends Vue {
 
   // Reverse the status of clientIdWithTime.
   private reverseClientIDWithTime() {
+    // when the client connected, we should disable this icon event
+    let curClientConnection: any = this.client
+    if (curClientConnection.connected === true) {
```
ysfscream, line 156:
```js
if (curClientConnection.connected) {}
```

It will be easier to write like this.

ysfscream:
> We expect that the form cannot be modified when connected, Now we cannot modify the form while connected.
> after

Disabled input seems a bit ugly. Maybe you can set the `input` background color to `transparent`, and see some. Then look at the effect and try not to destroy the original style, our purpose is just to disable it.

src/views/connections/ConnectionInfo.vue
```diff
@@ -26,28 +26,40 @@
...
               </a>
             </el-tooltip>
-            <el-input size="mini" v-model="connection.clientId">
-              <i slot="suffix" title="Refresh" class="el-icon-refresh" @click="refreshClientId"> </i>
+            <el-input size="mini" v-model="connection.clientId" :disabled="client.connected">
+              <i
+                slot="suffix"
+                title="Refresh"
+                class="el-icon-refresh"
+                :class="client.connected === true ? 'icon-oper-disable' : ''"
```
0721Betty, line 34: client.connected?

oceanlvr: Sorry, I will pay attention to this detail.

oceanlvr:
before

image

after

image

0721Betty (approved):

ysfscream (approved):

PR #433, opened 2020-12-25
feat(logs): Attach an empty log page when opening the app 

oceanlvr:
* Attach an empty log page when opening the app 
* Reformat code
* Resize icon


oceanlvr:
when we open `MQTTX`, the log system will create a new log file to record your operation instead of reading from the newest log file from `${APP_DATA_PATH}/logs/${newest_log_file}`

src/views/log/index.vue
```diff
@@ -85,9 +48,7 @@ export default class Script extends Vue {
       persistent: true,
     })
     watcher.on('add', () => {
-      const newestLog = this.getNewestLogFileName(LOG_PATH)
-      if (!newestLog) return
-      this.logValue = fs.readFileSync(path.join(LOG_PATH, newestLog)).toString()
+      this.logValue = fs.readFileSync(path.join(LOG_PATH, 'log')).toString()
```
ysfscream, line 51: Will there be a problem of reading large `log` files here? Then an error occurs or the reading is slow? My suggestion is to use asynchronous read:

```js
fs.readFile(path.join(LOG_PATH, 'log'), 'utf-8', (error, data) => {
	if (error) {
		// Handle error
	} else {
		this.logValue = data
	}
})
```

oceanlvr: THANKS FOR YOUR SUGGESTION!
You are right, I think we should update the log page by `buffer stream` instead of the whole file.
The solution now is to read the file stream directly into the page, and Obviously wasted a lot of IO time.

So I will fix it next commit, here are some references:

- [https://morning.work/page/2016-07/readable-stream-and-log-files-processing-in-nodejs.html](https://morning.work/page/2016-07/readable-stream-and-log-files-processing-in-nodejs.html)
- [https://zhuanlan.zhihu.com/p/76507612](https://zhuanlan.zhihu.com/p/76507612)

oceanlvr:
The next step needs to be done:

* Bury the log
* Optimize log stream

src/views/log/index.vue
ysfscream: OK, it's great! Then I merge this PR first.

0721Betty (approved):

ysfscream (approved):

PR #441, opened 2021-01-05
fix(logs): set logger for connectionDetail page

oceanlvr:
I set logger for the fellow situation

* Message publish success
* Message receive
* Script/Import/Export data status 
* MQTT.js status record

Parts not recorded as logs

* UI operation
* Some operation  just affect the view module data (such as filter out some data by condition)

### DIFF
--- a/src/background.ts
+++ b/src/background.ts
@@ -3,7 +3,7 @@
 import { app, protocol, BrowserWindow, ipcMain, shell, Menu } from 'electron'
 import { createProtocol, installVueDevtools } from 'vue-cli-plugin-electron-builder/lib'
 import { updateConnectionMessage } from '@/api/connection'
-import { getlogger, quitAndRenameLogger } from './utils/logger'
+import { getCustomLogger, quitAndRenameLogger } from './utils/logger'
 import db from './database/index'
 import updateChecker from './main/updateChecker'
 import getMenuTemplate from './main/getMenuTemplate'
@@ -87,7 +87,7 @@ function createWindow() {
     backgroundColor: theme === 'dark' ? '#232323' : '#ffffff',
     icon: `${__static}/app.ico`,
   })
-  const logger = getlogger('main thread', 'info')
+  const logger = getCustomLogger('main thread', 'info')
   logger.info('App init')
 
   // Menu Manger
--- a/src/components/SubscriptionsList.vue
+++ b/src/components/SubscriptionsList.vue
@@ -104,6 +104,8 @@ import LeftPanel from '@/components/LeftPanel.vue'
 import MyDialog from '@/components/MyDialog.vue'
 import { ConnectionModel, SubscribeErrorReason } from '../views/connections/types'
 import VueI18n from 'vue-i18n'
+import log4js from 'log4js'
+import { getCustomLogger } from '@/utils/logger'
 
 @Component({
   components: {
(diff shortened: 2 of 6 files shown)


oceanlvr:
I set some `TODO/FIXME`  TAGs,  I'm not sure what to do with them. I want to ask for your advice, Thanks!

src/views/connections/ConnectionsDetail.vue
```diff
@@ -282,6 +282,7 @@ import { ipcRenderer } from 'electron'
 import mqtt, { MqttClient, IClientOptions } from 'mqtt'
 import { v4 as uuidv4 } from 'uuid'
 import _ from 'lodash'
+import { getlogger } from '@/utils/logger'
```
ysfscream, line 285: Please use camel case to name, I ignored it in my previous review, sorry.

```diff
+ getLogger
- getlogger
```

src/views/connections/ConnectionsDetail.vue
```diff
@@ -510,13 +511,15 @@ export default class ConnectionsDetail extends Vue {
 
   // Connect
   public connect(): boolean | void {
+    const logger = getlogger('connectionDetail', 'debug')
```
ysfscream, line 514: I think it’s necessary to refactor here, repeat it many times.

src/views/connections/ConnectionsDetail.vue
```diff
@@ -541,13 +545,14 @@ export default class ConnectionsDetail extends Vue {
           this.$emit('delete')
           this.$message.success(this.$t('common.deleteSuccess') as string)
           this.removeActiveConnection({ id: res.id as string })
+          logger.info(`MQTTX remove connection ${res.id} success`)
```
ysfscream, line 548: This `id` has been hidden all the time, and it is impossible to accurately identify which connection it is. Maybe you should add the connection `name` to display.

src/views/connections/ConnectionsDetail.vue
```diff
@@ -775,7 +793,9 @@ export default class ConnectionsDetail extends Vue {
   }
   // Delete topic item
   private handleTopicDelete() {
+    const logger = getlogger('connectionDetail', 'debug')
     this.getMessages()
+    logger.info('Delete the connection topic')
```
ysfscream, line 798: We should be able to know which topic was deleted?

src/views/connections/ConnectionsDetail.vue
oceanlvr: I think so, I think maybe we can set one `logger` to one `page`.
Init a logger in the `create` event.
what your opinion?

ysfscream: Yes, in the `created` or set a `computed`.

src/background.ts
```diff
@@ -3,7 +3,7 @@
 import { app, protocol, BrowserWindow, ipcMain, shell, Menu } from 'electron'
 import { createProtocol, installVueDevtools } from 'vue-cli-plugin-electron-builder/lib'
 import { updateConnectionMessage } from '@/api/connection'
-import { getlogger, quitAndRenameLogger } from './utils/logger'
+import { getCustomLogger, quitAndRenameLogger } from './utils/logger'
```
oceanlvr, line 6: `./utils/logger` change to `@/utils/logger` maybe better

ysfscream (approved):

src/components/SubscriptionsList.vue
```diff
@@ -116,6 +118,10 @@ export default class SubscriptionsList extends Vue {
   @Prop({ required: true }) public connectionId!: string
   @Prop({ required: true }) public record!: ConnectionModel
   @Prop({ type: String, default: '60px' }) public top!: string
+  @Prop({
```
ysfscream, line 121: Why do you use Prop to initialize the `logger` default value?

ysfscream:
Maybe we can optimize like this, the logger method should be a widely used global function, We can expand into the form of plugins, add some Vue instance methods by attaching them to Vue.prototype.

main.js
```js
Vue.use(Logger)

new Vue({
  //... options
})
```

component.vue
```js
this.$logger.info('debug')
```

Then you can call logger function at will in each component, and we only need to initialize it once, before Vue is constructed.

ref:
https://vuejs.org/v2/guide/plugins.html#Writing-a-Plugin

Issue #427, opened 2020-12-21
When jumping from the creation page to the connection page, the editor cannot be initialized


ysfscream:
### Description

Go to other pages, click the `+` Create Connection button, and then go to the connection page, the editor cannot work.

ysfscream:
Fixed on https://github.com/emqx/MQTTX/releases/tag/v1.5.0

PR #446, opened 2021-01-11
fix(MsgPublish): fix cannot initialized editor jump from creation page

oceanlvr:
fix:#[427](https://github.com/emqx/MQTTX/issues/427)

### DIFF
--- a/src/components/MsgPublish.vue
+++ b/src/components/MsgPublish.vue
@@ -56,6 +56,7 @@ import Editor from '@/components/Editor.vue'
 import { MessageModel } from '../views/connections/types'
 import convertPayload from '@/utils/convertPayload'
 import { v4 as uuidv4 } from 'uuid'
+import { Route } from 'vue-router'
 
 @Component({
   components: {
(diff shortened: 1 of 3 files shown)


oceanlvr:
It seems works when I set a watcher to  the router changes:
success1
success2

src/components/MsgPublish.vue
```diff
@@ -115,13 +115,10 @@ export default class MsgPublish extends Vue {
...
     }
   }
-
-  private mounted() {
-    ipcRenderer.on('initEditor', () => {
-      const editorRef = this.$refs.payloadEditor as Editor
-      editorRef.initEditor()
-      ipcRenderer.removeAllListeners('initEditor')
-    })
+  @Watch('$route.params.id') private handleIdChanged() {
```
ysfscream, line 118: Add new and old value params in `handleIdChanged`? if the old value is `0`, we will need to initEditor?

ysfscream (approved):

0721Betty (approved):

PR #461, opened 2021-01-25
fix(UI): fixed the jitter of the input component

oceanlvr:
The old version of the input component will have a length when you click the plus and minus buttons, use el-input-number instead.


oceanlvr:
The current input component clicks to increase only once. 
image

0721Betty (approved):

ysfscream (approved):

PR #463, opened 2021-01-26
fix(feat): add connect loading icon when the ClientInfo scroll

oceanlvr:


src/views/connections/ConnectionsDetail.vue
```diff
@@ -61,6 +61,13 @@
                   </a>
                 </el-tooltip>
               </span>
+              <span v-if="!showClientInfo && this.connectLoading">
+                <el-tooltip placement="bottom" :effect="theme !== 'light' ? 'light' : 'dark'" :open-delay="500">
+                  <a class="connect-loading" href="javascript:;">
+                    <i class="el-icon-loading"></i>
+                  </a>
+                </el-tooltip>
```
ysfscream, line 69: Why do you need this `tooltip`?

oceanlvr: Ok, this icon will not be affected by the theme. I have fixed it.

PR #474, opened 2021-01-28
fix(connections): remove unused modules

ysfscream:


ysfscream:
@oceanlvr 
1. Removed some modules that were not used after import
2. Optimized the protocol type

src/types/global.d.ts
```diff
@@ -13,6 +13,17 @@ declare global {
...
+    ws = 'ws',
+    wss = 'wss',
+    mqtt = 'mqtt',
+    mqtts = 'mqtts',
+  }
+
+  type ProtocolMap = {
+    [key in ProtocolOption]: string
+  }
+
```
oceanlvr, line 26: 🆒 🆒 🆒 🆒 🆒

oceanlvr (approved):

PR #477, opened 2021-01-29
fix(connections): fix the problem that the client id cannot be refreshed

ysfscream:


src/components/MsgPublish.vue
```diff
@@ -32,15 +32,7 @@
...
       }"
     >
-      <Editor
-        ref="payloadEditor"
-        id="payload"
-        :lang="payloadLang"
-        v-model="msgRecord.payload"
-        @enter-event="send"
-        @focus="handleInputFoucs"
-        @blur="handleInputBlur"
```
oceanlvr, line 42: Why we Should delete those events and do we need to remove its `IPC` part in main?

oceanlvr (approved):

PR #518, opened 2021-02-18
fix(MsgPublish): fix editor error when router jump from other page to creation & connection

oceanlvr:
### PR Checklist

When we jump order by `other page` -> `creation page` -> `connection page`, and press on `ctrl/cmd` + `enter`, we will get two message.

I hold the view that it caused by the `editor` init early when we jump from `other page` to `creation page`, so we init two`editor`s here.

We should destroy the early init `editor`  component, like before we do: #446 

Here is my solution:

Spy on the router, when we jump from `other page` to `creation page`, we just destroy the editor which init in `MsgPublish`.

#### Issue Number

None:

When we follow jump order by `other page` -> `creation page` -> `connection page`, and press on `ctrl/cmd` + `enter`, we will get only one message.
success

- [ ] Yes
- [x] No

### DIFF
--- a/src/views/connections/ConnectionsDetail.vue
+++ b/src/views/connections/ConnectionsDetail.vue
@@ -368,6 +368,26 @@ export default class ConnectionsDetail extends Vue {
   @Getter('showClientInfo') private clientInfoVisibles!: { [id: string]: boolean }
   @Getter('currentScript') private scriptOption!: ScriptState | null
 
+  /**
+   * Notice: when we jump order by `other page` -> `creation page` -> `connection page`,
+   * MsgPublish/editor twice which is not we expected, it should be init only once.
+   * `other page` -> `creation page` the MsgPublish/editor will init, `creation page` -> `connection page` init editor again.
+   * So when route jump order by `other page` -> `creation page`, we need to operate editor creation and destroy manually by listening on route.
+   * relative PR: https://github.com/emqx/MQTTX/pull/518 https://github.com/emqx/MQTTX/pull/446
+   */
+  @Watch('$route.path', { immediate: true, deep: true })
+  private handleIdChanged(to: string, from: string) {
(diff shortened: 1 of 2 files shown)


oceanlvr:
> When we jump order by `other page` -> `creation page` -> `connection page`, and press on `ctrl/cmd` + `enter`, we will get two message.
I hold the view that it caused by the `editor` init early when we jump from `other page` to `creation page`, so we init two`editor`s here.
We should destroy the early init `editor`  component, like before we do: #446 
Here is my solution:
Spy on the router, when we jump from `other page` to `creation page`, we just destroy the editor which init in `MsgPublish`.

@ysfscream Do you think this is a good solution? :)

ysfscream:
I think you only need to listen for `$router.params.id` changes in `MsgPublish.vue` component?

ysfscream (approved):

src/views/connections/ConnectionsDetail.vue
```diff
@@ -294,7 +294,7 @@ import { ipcRenderer } from 'electron'
 import mqtt, { MqttClient, IClientOptions } from 'mqtt'
 import { v4 as uuidv4 } from 'uuid'
 import _ from 'lodash'
-
+import { Route } from 'vue-router'
```
ysfscream, line 297: Unused import

src/views/connections/ConnectionsDetail.vue
```diff
@@ -368,6 +368,16 @@ export default class ConnectionsDetail extends Vue {
   @Getter('showClientInfo') private clientInfoVisibles!: { [id: string]: boolean }
   @Getter('currentScript') private scriptOption!: ScriptState | null
 
+  @Watch('$route.path', { immediate: true, deep: true })
```
ysfscream, line 371: Add some comments

ysfscream (approved):

PR #550, opened 2021-04-22
fix(connecionsList): fix connection hightlight incorrect issue

oceanlvr:
### PR Checklist

highlight issue like:

1. click a connection `A`
2. click another connection `B`
3. connection current connection
4. the  highlight connection is `A`(expect highlight `B`)

#### Issue Number

None.

Lead to the `current-node-key` is not reactive, so it won't change like  we expect.

ref issue: https://github.com/ElemeFE/element/issues/17246

- [ ] Yes
- [x] No

#### Specific Instructions

### DIFF
--- a/src/views/connections/ConnectionsList.vue
+++ b/src/views/connections/ConnectionsList.vue
@@ -28,7 +28,6 @@
           node-key="id"
           highlight-current
           default-expand-all
-          :current-node-key="connectionId"
           @node-drop="handleDrop"
           @node-drag-end="handleDragEnd"
           :allow-drop="allowDrop"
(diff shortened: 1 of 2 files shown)


ysfscream (approved):

src/views/connections/ConnectionsList.vue
```diff
@@ -262,6 +268,14 @@ export default class ConnectionsList extends Vue {
   }
 
   private loadData() {
+    this.$nextTick(() => {
+      if (this.$route.params.id) {
+        const treeRef = this.$refs.tree as ElTree<ConnectionModelTree['id'], ConnectionModelTree>
+        treeRef.setCurrentKey(this.$route.params.id)
+        this.connectionId = this.$route.params.id
+      }
+    })
```
ysfscream, line 277:
```js
const { id } = this.$route.params
if (id) {
	treeRef.setCurrentKey(id)
	this.connectionId = id
}
```

src/views/connections/ConnectionsList.vue
```diff
@@ -304,8 +318,10 @@ export default class ConnectionsList extends Vue {
     if (!this.handleCollectionNameValidate(data.name)) {
       this.handleEditCancel(node, data)
     } else {
-      data.isEdit = false
-      this.flushCollectionChange()
+      if (data) {
+        data.isEdit = false
+        this.flushCollectionChange()
+      }
     }
```
ysfscream, line 325:
```js
else if (data) {} // Like this?
```

ysfscream (approved):

PR #554, opened 2021-04-30
fix(MsgPublish): support remove duplicates for payload & headers

oceanlvr:
### PR Checklist

#### Issue Number

None

support remove duplicates for payload & headers.

- [ ] Yes
- [x] No

#### Specific Instructions

### DIFF
--- a/src/components/MsgPublish.vue
+++ b/src/components/MsgPublish.vue
@@ -38,15 +38,21 @@
       <div class="publish-right-bar">
         <div class="history-icon">
           <el-button
-            :disabled="historyIndex === 0"
+            :disabled="historyIndex === (0 || -1)"
             circle
             size="mini"
             icon="el-icon-back"
             @click="decrease"
           ></el-button>
-          <el-button circle size="mini" icon="el-icon-minus" @click="back"></el-button>
           <el-button
-            :disabled="historyIndex === payloadsHistory.length - 1"
+            circle
+            :disabled="historyIndex === (payloadsHistory.length - 1 || -1)"
+            size="mini"
+            icon="el-icon-minus"
+            @click="back"
+          ></el-button>
+          <el-button
+            :disabled="historyIndex === (payloadsHistory.length - 1 || -1)"
             circle
             size="mini"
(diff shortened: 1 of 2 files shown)


src/utils/mqttUtils.ts
```diff
@@ -131,4 +133,25 @@ export const getMQTTProtocol = (data: ConnectionModel): Protocol => {
...
+}
+
+export const hasMessageHeader = async (data: HistoryMessageHeaderModel): Promise<boolean> => {
+  const headers = await loadHistoryMessageHeaders()
+  const res = headers.some((el: HistoryMessageHeaderModel) => {
+    return (
+      data.connectionId === el.connectionId &&
+      data.qos === el.qos &&
+      data.topic === el.topic &&
+      data.retain === el.retain
```
ysfscream, line 151: There are too many logical operations here, how about concatenating them as strings?

src/components/MsgPublish.vue
```diff
@@ -191,8 +198,8 @@ export default class MsgPublish extends Vue {
       topic: this.msgRecord.topic,
       qos: this.msgRecord.qos,
     } as HistoryMessageHeaderModel
-    await createHistoryMessagePayload(payload)
-    await createHistoryMessageHeader(header)
+    !(await hasMessagePayload(payload)) && (await createHistoryMessagePayload(payload))
+    !(await hasMessageHeader(header)) && (await createHistoryMessageHeader(header))
```
ysfscream, line 202: I am not very clear what is done here.

oceanlvr: I want to insert a payload message if it hasn't been inserted.

```javascript
const res = await hasMessagePayload(payload);
if (res) {
  await createHistoryMessagePayload(payload)
}
```

src/utils/mqttUtils.ts
oceanlvr: I think there should not concatenating as a single string because the `===` operator makes sure the type currently.

ysfscream (approved):

Issue #455, opened 2021-01-21
Save publish-topics historically


zyphlar:
Much like a browser window with URLs, being able to quickly recall and publish to a past topic name instead of having to re-type every time would be very useful. Maybe a combo drop-down text box?

ysfscream:
Hi, thanks for your support for `MQTT X`. This is a good suggestion worth discussing, there is already a plan that includes this feature #428 . The specific implementation may need to be redesigned, I will complete it as soon as possible.

oceanlvr:
@zyphlar Hi, the next version will have support for the quick selection of historical records.

PR #557, opened 2021-05-07
feat(Msgpublish): support select message header by selector

oceanlvr:
### PR Checklist

None.

#### Issue Number

https://github.com/emqx/MQTTX/issues/455

support quick select message header:

image

check one option it will be auto-complete the `payload` `topic` `retain` options.

and when we send a message with a new header, this option will be added to the selector item.(max item number is 10 now).

- [ ] Yes
- [x] No

#### Specific Instructions

### DIFF
--- a/src/components/MsgPublish.vue
+++ b/src/components/MsgPublish.vue
@@ -25,6 +25,21 @@
         @blur="handleInputBlur"
       >
       </el-input>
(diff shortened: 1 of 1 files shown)


oceanlvr:
@ysfscream PTAL, how about this design?

ysfscream:
> @ysfscream PTAL, how about this design? :)

YES! It's GOOD.

ysfscream (approved):

PR #628, opened 2021-07-08
feat(config): add support for mac universal arch

m01i0ng:
### PR Checklist

image

- [ ] Yes
- [x] No

#### Specific Instructions


ysfscream:
@m01i0ng Hi! Thank you for submitting your PR! Can I ask a question? Is this a modification that can be packaged with a generic chip architecture installer? Including Apple M1 and Inter CPU?

m01i0ng:
> @m01i0ng Hi! Thank you for submitting your PR! May I ask a question? Is this a modification that can be packaged with a generic chip architecture installer? Including Apple M1 and Inter CPU?

Sure! I've tried with an intel mac and everything ok.
截屏2021-07-09 上午11 59 45

m01i0ng:
Mac universal app contains both two platform's runtime so the bundle size will increase by 100M+.

ysfscream:
YES! 100M+ is a little too big. 😂

OK! Anyway I will merge it! thanks!

ysfscream (approved):

package.json
```diff
@@ -76,8 +76,8 @@
     "husky": "^4.3.0",
     "lint-staged": "^10.4.2",
     "monaco-editor-webpack-plugin": "^4.0.0",
-    "node-sass": "^4.13.1",
     "prettier": "^2.0.5",
+    "sass": "^1.35.2",
```
ysfscream, line 80: After this dependency change, I see these warnings in the console when starting, is there a problem?

image

ysfscream: @m01i0ng I updated `element-ui` and `sass-loader` versions, but still have this problem

m01i0ng: Element-ui use node-sass and I replaced it with `dart-sass` because `node-sass` can't be compiled with mac arm arch. I saw `dart-sass` deprecate `/` in `1.33.0`, so `yarn add sass@~1.32.13 -D` will remove the warnings which is a previous latest version.

ysfscream: OK. Thanks a lot. To ensure both compiled with mac arm arch and the removal of syntax warnings or to use the 1.32 version for now, I will update it.

ysfscream: And I will build the universal version and upload the mac assets to the release 1.6.0

ysfscream:
@m01i0ng Hi! I am building the mac universal packages now, but I got this error when packaging. Is it a problem with my Node version? Which Node version are you using?

image


Node: v14.5.0

m01i0ng:
@ysfscream Hi! According to nodejs's changlog, v16 is the first version that support m1, so I guess node v16 is required. And I use v16.4.0 now.
https://github.com/nodejs/node/blob/b7136827f5d9f4f65bd0a60e75c4a9531cf4312a/doc/changelogs/CHANGELOG_V16.md#toolchain-and-compiler-upgrades

m01i0ng:
@ysfscream v16 is not required. `fs.rm` was added in v14.14.0, so the latest v14 works fine.

ysfscream:
@m01i0ng YES! Thanks for the answer! I upgraded the node version and fixed the issue, now the universal package is uploaded to the release assets and the release note is updated. https://github.com/emqx/MQTTX/releases/tag/v1.6.0

Thanks again! 👏

m01i0ng:
> @m01i0ng YES! Thanks for the answer! I upgraded the node version and fixed the issue, now the universal package is uploaded to the release assets and the release note is updated. https://github.com/emqx/MQTTX/releases/tag/v1.6.0
> 
> 
> 
> Thanks again! 👏

Nice! And don't forget brew's cask. I think we can add a judgement, if m1, gives the universal bundle and if intel then the normal one.

ysfscream:
@m01i0ng Good one. That's a great suggestion!

PR #636, opened 2021-07-20
chore(deps): add sqlite & typeorm

ysfscream:
### PR Checklist

None

#### Issue Number

Example: None

Add SQLite database and TypeORM ORM library.

- [ ] Yes
- [x] No

#### Specific Instructions


src/database/sqlite.ts
```diff
@@ -0,0 +1,48 @@
...
+  constructor() {
+    this.init()
+  }
+  public async init(): Promise<void> {
+    this.connection = await createConnection({
+      type: 'sqlite',
+      database: path.join(STORE_PATH, 'MQTTX.db'),
+      entities: [Connections],
+    })
+    if (this.connection.isConnected) {
```
oceanlvr, line 33: maybe it could be put into `package.json`, and we could use script like `typeorm schema:sync`,to update the schema. 
In this time, maybe we could write some schema `database update log` file, so we could easily trace the data schema.

oceanlvr: > This command will execute schema synchronization. Note, to make command line tools work, you must create an ormconfig.json file.

is these config could be manage by `ormconfig.json`, maybe we could put in there.

oceanlvr:
I still wan't to add more thing in this layer, such as `hooks`.

like this :

```ts
const { loading, useFetchConnection, errorHandler } = useFetchConnectionHooks({
    id: <id>,
})

await useFetchConnection() // sync
useFetchConnection() //async


<div v-if="loading" >

try {
  useFetchConnection()
} catch(err){
  errorHandler(err)
}

```

src/database/sqlite.ts
ysfscream: YES! I will add ormconfig.json.

oceanlvr (approved):

ysfscream:
@oceanlvr I refactor the use SQLite function and dropped the class methods. If there are no problems, please merge it and test.

PR #641, opened 2021-07-23
feat(typeorm): add typeorm CLI and refactor typeorm config

oceanlvr:
### PR Checklist

dup config `orm` file.

#### Issue Number

None.

refactor `orm` config file, add CLI command. and document.

- [ ] Yes
- [x] No

#### Specific Instructions


package.json
```diff
@@ -22,6 +29,7 @@
   },
   "main": "background.js",
   "dependencies": {
+    "appdata-path": "^1.0.0",
```
oceanlvr, line 32: I use this to replace `getPath('userData')`, because this is a `Node` environment. And both orm config files, which make problem solved.

src/database/README.md
```diff
@@ -0,0 +1,49 @@
...
+
+1. `yarn run db:log` check CLI sync execute SQL queries
+2. `yarn run db:diagram` get a latest ER diagram
+3. `db:migration:sync` sync models to database
+
+Otherwise, Please follow bellow steps and make sure what you do. Or there is a high risk to lost data.
+
+1. `yarn run db:log` check CLI sync execute SQL queries
+2. `db:migration:generate -n {migration_name}` Automatic migration generate
+3. `db:migration:run` make pending migrations executed. NOTICE: you should execute this code before the APP open since the table model changed.
```
oceanlvr, line 39: There might be use by update code in production mode, but not directly run CLI `db:migration:run` command.

tsconfig.commonjs.json
```diff
@@ -0,0 +1,6 @@
+{
+  "extends": "./tsconfig.json",
```
oceanlvr, line 2: import this file to solve the `commonjs` require problem which leads by typeorm CLI.

oceanlvr:
@ysfscream PTAL

oceanlvr:
database

src/database/README.md
```diff
@@ -0,0 +1,49 @@
+# Database Migration
```
ysfscream, line 1: Please move the `Database Migration` to `Contributing Guide`, because no one wants to care about how the database works in the README.

If there is a command that is important and necessary, please write it in the `Develop` module of the README.

ysfscream:
@oceanlvr Please remove the `src/database/database.png`. This is a PNG file that needs to be referenced during development and does not need to be in the src directory.

src/database/models/ConnectionEntity.ts
```diff
@@ -0,0 +1,91 @@
+// see https://typeorm.io/#/entities/column-types-for-sqlite--cordova--react-native--expo
+import { Entity, Column, PrimaryGeneratedColumn, OneToMany, ManyToOne } from 'typeorm'
+import MessageEntity from './MessageEntity'
+import SubscriptionEntity from './SubscriptionEntity'
+import CollectionEntity from './CollectionEntity'
```
ysfscream, line 5: A blank line here.

src/database/models/ConnectionEntity.ts
```diff
@@ -0,0 +1,91 @@
...
+  clean!: boolean
+
+  @Column({ type: 'varchar' })
+  protocol?: string
+
+  @Column({ type: 'varchar' })
+  host?: string
+
+  @Column({ type: 'integer' })
+  port?: number
```
ysfscream, line 27: These are not optional, they are required. `?` -> `!`

src/database/models/ConnectionEntity.ts
```diff
@@ -0,0 +1,91 @@
...
+
+  @Column({ type: 'integer', default: 60 })
+  keepalive?: number
+
+  @Column({ type: 'integer' })
+  connectTimeout?: number
+
+  @Column({ type: 'boolean' })
+  reconnect?: boolean
```
ysfscream, line 36: These are not optional, they are required. `?` -> `!`

ormconfig.ts
```diff
@@ -0,0 +1,49 @@
...
+import getAppDataPath from 'appdata-path'
+
+import ConnectionEntity from './src/database/models/ConnectionEntity'
+import MessageEntity from './src/database/models/MessageEntity'
+import SubscriptionEntity from './src/database/models/SubscriptionEntity'
+import ScriptEntity from './src/database/models/ScriptEntity'
+import SettingEntity from './src/database/models/SettingEntity'
+import CollectionEntity from './src/database/models/CollectionEntity'
+import HistoryMessageHeaderEntity from './src/database/models/HistoryMessageHeaderEntity'
+import HistoryMessagePayloadEntity from './src/database/models/HistoryMessagePayloadEntity'
```
ysfscream, line 12: I'm a little curious, how does it work here?

oceanlvr: https://github.com/tc39/proposal-decorators

maybe the entity use `decorators` to declare information that typeorm needed.

src/database/README.md
oceanlvr: OK, I will put it into `Contributing Guide#Develop`

oceanlvr:
> @oceanlvr Please remove the `src/database/database.png`. This is a PNG file that needs to be referenced during development and does not need to be in the src directory.

Yes, it is only used on dev mode.

But I think that picture is necessary, which makes the developer clear knows the current database model.

And best practice I think is to keep a `CHANGELOG.md` record the migration of the database.

those files will not increase the size of those final packages.

src/database/models/ConnectionEntity.ts
oceanlvr: Yes, but I found those `!/?` symbols will not affect the data model.

It may just affect Typescript with UI type.

If we want to define a nullable type in the database.

We should use

```typescript
  @Column({ type: 'varchar', nullable: false })
  name!: string
```
or nullable like this

```typescript
  @Column({ type: 'varchar', nullable: true })
  name?: string
```

for `nullable` it is false for default

ormconfig.ts
oceanlvr: I think those entities are just some object with decorators information, and typeorm generate a class for database model

src/database/models/ConnectionEntity.ts
```diff
@@ -0,0 +1,91 @@
...
+  ssl?: boolean
+
+  @Column({ type: 'varchar' })
+  mqttVersion?: string
+
+  @Column({ type: 'number' })
+  unreadMessageCount?: number
+
+  // @Column()
+  // client!: string//TODO
```
ysfscream, line 60: Create the Client table

```
+-------------+--------------+----------------------------+
|                        Client                           |
+-------------+--------------+----------------------------+
| id          | int(11)      | PRIMARY KEY AUTO_INCREMENT |
| connected   | boolean      |                            |
+-------------+--------------+----------------------------+
```

Set up a one-to-one relationship

ConnectionEntity.ts

```js
Class ConnectionEntity {
	@OneToOne(() => ClientEntity)
    @JoinColumn()
   client: ClientEntity;
}
```

ysfscream: OK, this table is unused. Remove it!

src/database/models/ConnectionEntity.ts
```diff
@@ -0,0 +1,91 @@
...
+
+  // @Column()
+  // will!: string //TODO
```
ysfscream, line 63: Set up ManyToMany relationships.

ysfscream: Create the `will` table.

ysfscream: Sorry, I think this is not a ManyToMany relationship.

One connection only has one `will` instance, and one will data only belongs to the one connection.

```js
class Connection Entity {
	@OneToOne(() => WillEntity)
    @JoinColumn()
	will?: WillEntity
}
```

ysfscream: ref: https://typeorm.io/#/one-to-one-relations

ysfscream (approved):

ysfscream:
Nice work!

PR #646, opened 2021-07-29
migration(db): init table models

oceanlvr:
### PR Checklist

init table models use by `migration` script

#### Issue Number

None.

init tables

- [x] Yes
- [ ] No

#### Specific Instructions

### DIFF
--- a/src/background.ts
+++ b/src/background.ts
@@ -38,10 +38,7 @@ protocol.registerSchemesAsPrivileged([{ scheme: 'app', privileges: { secure: tru
 
 app.allowRendererProcessReuse = false
 
-let { ConnectionInit, ConnectionDestory } = useConnection({
-  doMigrations: false,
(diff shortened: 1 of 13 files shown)


src/background.ts
```diff
@@ -38,10 +38,7 @@ protocol.registerSchemesAsPrivileged([{ scheme: 'app', privileges: { secure: tru
 
 app.allowRendererProcessReuse = false
 
-let { ConnectionInit, ConnectionDestory } = useConnection({
-  doMigrations: false,
-  undoMigrations: false,
-} as initOptionModel)
+let { ConnectionInit, ConnectionDestory } = useConnection()
```
ysfscream, line 41: Why do you use the `let` here?

oceanlvr: sorry for mistake QAQ

ysfscream (approved):

PR #649, opened 2021-08-02
refactor(connection): migration connection api to service

oceanlvr:
### PR Checklist

refactor connection service API

- [ ] Yes
- [x] No

#### Specific Instructions

### DIFF
--- a/src/database/models/CollectionEntity.ts
+++ b/src/database/models/CollectionEntity.ts
@@ -1,4 +1,4 @@
-import { Entity, PrimaryGeneratedColumn, Column, OneToMany } from 'typeorm'
+import { Entity, PrimaryGeneratedColumn, Column, OneToMany, ManyToOne } from 'typeorm'
 import ConnectionEntity from './ConnectionEntity'
 
 @Entity('CollectionEntity')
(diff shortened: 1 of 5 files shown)


src/database/services/ConnectionService.ts
```diff
@@ -2,35 +2,137 @@ import { Service } from 'typedi'
...
+
+  public async loadAllConnectionsIds(): Promise<number[] | undefined> {
+    let res: number[] | undefined = undefined
+    const query: ConnectionEntity[] | undefined = await this.connectionRepository.find({
+      select: ['id'],
+    })
+    if (!query) {
+      return
     }
+    res = query.map((e) => e.id as number)
```
ysfscream, line 135: Why is it asserted here that the id is number?

src/database/services/ConnectionService.ts
```diff
@@ -2,35 +2,137 @@ import { Service } from 'typedi'
...
+  public async updateConnection(id: number, data: Partial<ConnectionModel>): Promise<ConnectionModel | undefined> {
+    let res: ConnectionEntity | undefined = await this.connectionRepository.findOne(id)
+    if (!res) {
+      return
+    }
+    deepMerge(res, data)
+    const query: ConnectionEntity | undefined = await this.connectionRepository.save(res)
+    return {
+      ...query,
+      isCollection: false,
```
ysfscream, line 44: Isn't the isCollection field inside ConnectionEntity?

src/database/services/ConnectionService.ts
```diff
@@ -2,35 +2,137 @@ import { Service } from 'typedi'
...
+    } as ConnectionModel
+    return res
+  }
+
+  public async loadAllConnections(): Promise<ConnectionModel[] | undefined> {
+    const query: ConnectionEntity[] | undefined = await this.connectionRepository.find()
+    if (!query) {
+      return
+    }
+    return query.map((e) => {
```
ysfscream, line 95: It is better to try not to use single letters for parameter naming.

src/database/services/ConnectionService.ts
oceanlvr: Yes, do we need to define it in ConnectionEntity?

ysfscream: This field is valid, right? If it's available, it's still stored in the database.

src/database/models/CollectionEntity.ts
```diff
@@ -7,17 +7,23 @@ export default class CollectionEntity {
...
-  payloadType!: string
-
-  @Column({ type: 'integer', comment: 'order in the collection' })
+  @Column({ type: 'integer', nullable: true, comment: 'order in the collection' })
   orderId!: number
 
-  @OneToMany(() => ConnectionEntity, (connection) => connection.collection)
-  connections!: ConnectionEntity[]
+  @Column({ type: 'boolean', default: true })
+  isCollection?: true
```
ysfscream, line 16: true -> boolean?

oceanlvr: maybe it's better just use true, so that we don't need to handle the type transfer in the services layer

ysfscream (approved):

PR #661, opened 2021-08-18
refactor(service): refactor connection & history service

oceanlvr:
### PR Checklist

None.

#### Issue Number

Example: None

- [x] refactor connection services include cascade operation
- [x] refactor history payload & header service
- [x] replace history payload & header API
- [ ] refactor & replace message service API
- [ ] refactor & replace subscription service API

- [ ] Yes
- [x] No

#### Specific Instructions

### DIFF
--- a/src/api/connection.ts
+++ b/src/api/connection.ts
@@ -8,29 +8,10 @@ export const loadConnections = (): ConnectionModel[] | [] => {
   return db.get<ConnectionModel[] | []>('connections')
 }
 
-export const loadSuggestConnections = (): ConnectionModel[] | [] => {
-  return db.get<ConnectionModel[] | []>('suggestConnections')
-}
-
-export const createSuggestConnection = (data: ConnectionModel): ConnectionModel => {
-  const suggestConnections = loadSuggestConnections()
-  if (suggestConnections.length > 9) {
-    const deleteId = suggestConnections[0].id
-    if (deleteId !== undefined) {
-      deleteSuggestConnection(deleteId.toString())
-    }
-  }
-  return db.insert<ConnectionModel>('suggestConnections', data)
-}
-
-export const deleteSuggestConnection = (id: string): ConnectionModel => {
-  return db.remove<ConnectionModel>('suggestConnections', id)
-}
-
 export const loadAllConnectionsIds = async (type: 'connections' | 'suggestConnections'): Promise<string[]> => {
   const connectionsIds: string[] = []
   let allConnections: ConnectionModel[]
-  allConnections = type === 'connections' ? await loadConnections() : await loadSuggestConnections()
+  allConnections = await loadConnections()
   allConnections.forEach((connection: ConnectionModel) => {
     if (connection.id) {
       connectionsIds.push(connection.id.toString())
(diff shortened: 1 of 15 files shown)


tsconfig.json
```diff
@@ -1,6 +1,6 @@
 {
   "compilerOptions": {
-    "target": "esnext",
+    "target": "es6",
```
oceanlvr, line 3: @ysfscream  this fixes the `?.` syntax issue, but I'm not sure this is will affect something?

src/views/connections/ConnectionsList.vue
```diff
@@ -223,8 +223,9 @@ export default class ConnectionsList extends Vue {
   private async flushCollectionChange() {
     const tree = _.cloneDeep(this.treeData)
     const { collectionService } = useServices()
+    // BUG: tree set bug -> can't remove collection corrrectly
```
oceanlvr, line 226: this is a known bug, which leads to the collection can't be deleted because the database should not use `set` (but lowdb is a nested database that works).
I'll fix it in next PR.

tsconfig.json
ysfscream: > To just answer the question, the difference is that import() expressions are understood in esnext, but not in es2015 (and as of TypeScript 2.9, import.meta is only understood with esnext).
I guess you could possibly expect an es2019 module target for both.

ref: https://github.com/microsoft/TypeScript/issues/24082

Maybe you could try the ES2019 or ES2020 ?

ysfscream: https://github.com/microsoft/TypeScript/wiki/Node-Target-Mapping

src/views/connections/ConnectionsList.vue
ysfscream: OK

src/database/services/ConnectionService.ts
```diff
@@ -42,60 +67,112 @@ export default class ConnectionService {
...
+      } else {
+        // no will relation in database
+        res.will = await this.willRepository.save({
+          lastWillPayload,
+          lastWillTopic,
+          lastWillQos,
+          lastWillRetain,
+          contentType: properties.contentType ? properties.contentType : '',
+        } as WillEntity)
+      }
```
ysfscream, line 115: These two pieces of code are a bit confusing.

Why can't we construct the final data by if and then save it only once?

```javascript
const data = { ... }
if (id) {
	data = { id, ...data }
}
willRepository.save(data)
```

src/api/connection.ts
```diff
@@ -133,11 +39,11 @@ export const importConnections = (data: ConnectionModel[]): Promise<string> => {
...
-            if (res.indexOf(id.toString()) === -1) {
-              createConnection(item)
-            } else {
-              updateConnection(id.toString() as string, item)
-            }
+            // if (res.indexOf(id.toString()) === -1) {
+            //   createConnection(item)
+            // } else {
+            //   updateConnection(id.toString() as string, item)
+            // }
```
ysfscream, line 46: Why not delete it? This connection api file should not be needed anymore?

tsconfig.json
oceanlvr: > https://github.com/microsoft/TypeScript/wiki/Node-Target-Mapping

AWESOME!!!

src/database/services/ConnectionService.ts
oceanlvr: because the will model is `one-to-one` relation with the connection model.

And will the model be none, so we need to handle this by hand.

This is a polyfill so that ensure every will has its connection, vice versa.

src/api/connection.ts
oceanlvr: This is a temporary solution. I'll fix later.

tsconfig.json
oceanlvr: > https://github.com/microsoft/TypeScript/wiki/Node-Target-Mapping
ES2019 works for node 12. awesome manual.

src/database/services/ConnectionService.ts
oceanlvr: const {
        id,
        properties = {
          contentType: '',
        },
        lastWillPayload = '',
        lastWillTopic = '',
        lastWillQos = 0,
        lastWillRetain = false,
      } = res.will
because we assign the default will model, and I think although this is a obect desctructor(ugly maybe?), but it clearly told that we should care the nested will model flat.

see more in the diff between `global.ts` WillModel and `WillEntity.ts` WillEntity.

ysfscream: But it has nothing to do with what I said here if you judge the save, right?

ysfscream (approved):

PR #662, opened 2021-08-20
refactor(service): refactor message services

oceanlvr:
### PR Checklist

#### Issue Number

Example: \#123

- [ ] Yes
- [ ] No

#### Specific Instructions


oceanlvr:
@ysfscream All refactor and API replace has finished! PTAL

src/background.ts
```diff
@@ -60,9 +59,10 @@ function handleIpcMessages() {
       newWindow(id, { isMac, theme, static: __static, path: '/new_window' })
     }
   })
-  ipcMain.on('saveMessages', (event: Electron.IpcMainEvent, ...args: any[]) => {
+  ipcMain.on('saveMessages', async (event: Electron.IpcMainEvent, ...args: any[]) => {
     const { id, receivedMessage } = args[0]
-    updateConnectionMessage(id, { ...receivedMessage })
+    const { messageService } = useServices()
+    await messageService.pushToConnection({ ...receivedMessage }, id)
```
ysfscream, line 65: How about we remove this logic in the main process and put it in the rendering process? There were other reasons to put it here at the time, but I don't think it's needed now, it might be a bit complicated.

src/router/index.ts
```diff
@@ -24,9 +24,12 @@ Router.prototype.push = function push(location: string) {
   return callRes.catch((err: Error) => err)
 }
 
-router.beforeEach((to, from, next) => {
+router.beforeEach(async (to, from, next) => {
   if (to.name === 'Connections') {
-    const connections: ConnectionModel[] | [] = loadConnections() || []
+    const { connectionService } = useServices()
+    // TODO: Need more test
+    // maybe we should not use async inside the foreach loop
```
ysfscream, line 31: There is the beforeEach hook function. Where is the forEach loop?

ysfscream:
The above are some minor issues, I will merge first and then test and discuss.

ysfscream (approved):

PR #683, opened 2021-09-09
fix(subscription): fix subscription remove failed issue

oceanlvr:
### PR Checklist

#### Issue Number

Example: \#123

- [ ] Yes
- [ ] No

#### Specific Instructions

### DIFF
--- a/src/database/services/ConnectionService.ts
+++ b/src/database/services/ConnectionService.ts
@@ -40,6 +40,8 @@ export default class ConnectionService {
   private entityToModel(data: ConnectionEntity): ConnectionModel {
     return {
       ...data,
+      messages: data.messages ?? [],
+      subscriptions: data.subscriptions ?? [],
       will: {
         ...data.will,
         properties: {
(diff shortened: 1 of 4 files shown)


src/database/services/ConnectionService.ts
```diff
@@ -299,4 +301,18 @@ export default class ConnectionService {
...
+      connectionId,
+    })
+    if (!query || !Array.isArray(query) || !query.length) {
+      await this.subscriptionRepository.save(subs)
+      return
+    } else {
+      await this.subscriptionRepository.remove(query)
+      await this.subscriptionRepository.save(subs)
+      return
+    }
```
ysfscream, line 316: return is used in the if statement, there is no need to use else. It's a little hard to understand why you need to remove and then save

ysfscream: But there are still problems with the syntax.

src/database/services/ConnectionService.ts
```diff
@@ -299,4 +301,17 @@ export default class ConnectionService {
...
+    const query: SubscriptionEntity[] = await this.subscriptionRepository.find({
+      connectionId,
+    })
+    if (!query || !Array.isArray(query) || !query.length) {
+      await this.subscriptionRepository.save(subs)
+      return
+    }
+    await this.subscriptionRepository.remove(query)
+    await this.subscriptionRepository.save(subs)
+    return
```
ysfscream, line 315: This return is useless.

ysfscream (approved):

PR #685, opened 2021-09-12
feat(messages): ⚡ improve message list performence

oceanlvr:
### PR Checklist

a lot of raw `push` operations in the short time lead to lag UI performance, use `vue-virtual-scroller` improve list render performance.

#### Issue Number

None.

better performance in the message list component.

- [ ] Yes
- [x] No

#### Specific Instructions

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -4,6 +4,7 @@ import VueI18n from 'vue-i18n'
 import VueClipboard from 'vue-clipboard2'
 import path from 'path'
 import log4js from 'log4js'
+import VueVirtualScroller from 'vue-virtual-scroller'
 
 import 'element-ui/lib/theme-chalk/index.css'
 import ElementLocale from 'element-ui/lib/locale'
@@ -34,6 +35,7 @@ ConnectionInit({ doMigrations: false, undoMigrations: false } as initOptionModel
   Vue.use(VueI18n)
   Vue.use(VueClipboard)
   Vue.use(VueLog4js, config)
+  Vue.use(VueVirtualScroller)
 
   const locale: Language = store.state.app.currentLang
   const vueI18n: VueI18n = new VueI18n({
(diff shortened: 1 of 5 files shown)


oceanlvr:
This commit may improve long list UI performance when the list length is less than `5000` items (with filter way). 
Maybe I should do more code review on the message split part, remove some array copy, etc.

src/main.ts
```diff
@@ -4,6 +4,7 @@ import VueI18n from 'vue-i18n'
 import VueClipboard from 'vue-clipboard2'
 import path from 'path'
 import log4js from 'log4js'
+import VueVirtualScroller from 'vue-virtual-scroller'
```
ysfscream, line 7: Perhaps this is the part where we can implement import in the component?

oceanlvr: we need this to install `VueVirtualScroller` plugin.

src/main.ts
```diff
@@ -34,6 +35,7 @@ ConnectionInit({ doMigrations: false, undoMigrations: false } as initOptionModel
   Vue.use(VueI18n)
   Vue.use(VueClipboard)
   Vue.use(VueLog4js, config)
+  Vue.use(VueVirtualScroller)
```
oceanlvr, line 38: we need to import `VueVirtualScroller` in main.ts for Vue plugin install

src/main.ts
oceanlvr: https://github.com/emqx/MQTTX/pull/685/files/0e36358040c9d89a5d937001dc1a6c7811b39789#diff-4fab5baaca5c14d2de62d8d2fceef376ddddcc8e9509d86cfa5643f51b89ce3dR38

ysfscream (approved):

ysfscream:
image

The topic tag is missing. It should be a scrollbar width issue.

Issue #705, opened 2021-09-24
[Bug] Data format conversion error


ysfscream:
#### What did I do

Enter the binary data, MQTTX exit, and then reopen.

#### What happened

The binary data is reconverted once more.

#### Expected

Binary data format remains unchanged

#### Environment

- OS: OS version, e.g. MacOS Big Sur 11.1
- MQTTX version, e.g. v1.6.0

#### More detail

More detail include `log` information or screenshot

![image](https://user-images.githubusercontent.com/21299158/134632836-12fd43d5-0bf7-4253-996d-40d5a4236718.png)

PR #707, opened 2021-09-24
fix(msgPublish): Data format conversion error

lileiemq:
#### Issue Number

#705 

Enter the binary data, MQTTX exit, and then reopen. Binary data format remains unchanged
image

- [ ] No

#### Specific Instructions

The cause of the problem is that it is already binary when it is opened, and it is caused by re-encoding.
Change the data type before obtaining the historical information, so that the first entry will not be re-encoded

### DIFF
--- a/src/components/MsgPublish.vue
+++ b/src/components/MsgPublish.vue
@@ -243,6 +243,9 @@ export default class MsgPublish extends Vue {
     const { historyMessageHeaderService, historyMessagePayloadService } = useServices()
     const headersHistory = (await historyMessageHeaderService.getAll()) ?? []
     const payloadsHistory = (await historyMessagePayloadService.getAll()) ?? []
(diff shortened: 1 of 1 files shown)


src/components/MsgPublish.vue
```diff
@@ -243,6 +243,9 @@ export default class MsgPublish extends Vue {
     const { historyMessageHeaderService, historyMessagePayloadService } = useServices()
     const headersHistory = (await historyMessageHeaderService.getAll()) ?? []
     const payloadsHistory = (await historyMessagePayloadService.getAll()) ?? []
+    if (payloadsHistory[payloadsHistory.length - 1]) {
```
ysfscream, line 246: `payloadsHistory.length - 1` Suggest changing to a variable, the code here is too long to read

oceanlvr (approved):

PR #708, opened 2021-09-25
fix(subscribe): fix retain message flash/mis-render issue

oceanlvr:
### PR Checklist

#### Issue Number

Example: \#123

image

I think the reason is the sync problem between connection / subscribe  (resubscribe) /  messageArrive event.

so delay subscribing (resubscribe), maybe a better way, which makes the sub-event behind the on message arrives.

- [ ] Yes
- [x] No

#### Specific Instructions

### DIFF
--- a/src/components/SubscriptionsList.vue
+++ b/src/components/SubscriptionsList.vue
@@ -253,7 +253,10 @@ export default class SubscriptionsList extends Vue {
       this.subRecord.topic = sub.topic
       this.subRecord.qos = sub.qos
       this.subRecord.color = sub.color
-      this.subscribe(this.subRecord)
+      // TODO sync subscribe with connection without setTimeout maybe better.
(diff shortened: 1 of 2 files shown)


ysfscream (approved):

src/components/SubscriptionsList.vue
```diff
@@ -253,7 +253,10 @@ export default class SubscriptionsList extends Vue {
       this.subRecord.topic = sub.topic
       this.subRecord.qos = sub.qos
       this.subRecord.color = sub.color
-      this.subscribe(this.subRecord)
+      // TODO sync subscribe with connection without setTimeout maybe better.
+      setTimeout(() => {
+        this.subscribe(this.subRecord)
+      }, 300)
```
ysfscream, line 259: Why use setTimeout?

PR #711, opened 2021-09-26
fix(msgPublish): payload changed after send msg

lileiemq:
#### Issue Number

Example: \#705

- [ ] No

#### Specific Instructions

### DIFF
--- a/src/components/MsgPublish.vue
+++ b/src/components/MsgPublish.vue
@@ -239,12 +239,12 @@ export default class MsgPublish extends Vue {
     ipcRenderer.removeAllListeners('sendPayload')
   }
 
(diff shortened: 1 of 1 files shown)


oceanlvr (commented):
https://github.com/emqx/MQTTX/blob/498f356ed3e4d96eba9263477ee84c708ba93866/src/views/connections/ConnectionsDetail.vue#L1136

Have you test this change would affect the `emit` function at `ConnectionsDetail.vue@ sendMessage`?

lileiemq:
When sending a message, isLoadData is undefined, which will not affect the sending of the message, nor will it affect the isNewPayload parameter

oceanlvr (approved):

ysfscream (approved):

PR #715, opened 2021-09-28
fix(connection): delete collection reload bug

lileiemq:

#### Issue Number

Example: no

delete group reload

- [ ] No

#### Specific Instructions

fix delete group will not reload

#### 
图片
Other information

### DIFF
--- a/src/views/connections/ConnectionsList.vue
+++ b/src/views/connections/ConnectionsList.vue
@@ -561,6 +561,7 @@ export default class ConnectionsList extends Vue {
       // delete the chilren
       const { collectionService } = useServices()
       await collectionService.delete(selectedCollection.id)
+      this.$emit('reload')
     }
 
     this.showCollectionsContextmenu = false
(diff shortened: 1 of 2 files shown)


oceanlvr (approved):
I understand the reproduce problem path, and it's a great issue finding, maybe I should do more tests. Thank you.

oceanlvr:
Do we need to add float tips like `Delete the including connection?` texture?

If the user chooses with `true` condition, we delete bellow connection/collection, etc.

else (false condition) we just move the collection's children to the top layout of tree. 

@lileiemq @ysfscream

ysfscream:
@oceanlvr 

> Do we need to add float tips like `Delete the including connection?` texture when we delete a `collection`?
> 
> If the user chooses with `true` condition, we delete bellow connection/collection, etc.
> 
> else (false condition) we just move the collection's children to the top layout of the Connection / Collection tree.
> 
> @lileiemq @ysfscream

It is a good suggestion to create a new issue track, but not to enter this version.

ysfscream (approved):

PR #716, opened 2021-09-29
fix(connection): reopen client topic not lasted

lileiemq:
#### Issue Number

Example: no

- [ ] No

#### Specific Instructions
Questions raised by colleagues after use。

After using topic to send data for '111', close the client and reopen it. The topic will become '555', which should be '111'

I can only think of deleting first and then adding, but I didn't think of a good solution

image
image

### DIFF
--- a/src/database/services/HistoryMessageHeaderService.ts
+++ b/src/database/services/HistoryMessageHeaderService.ts
@@ -38,6 +38,14 @@ export default class HistoryMessageHeaderService {
     return await this.messageRepository.save(data)
   }
 
+  public async orderChange(
+    id: string,
+    data: HistoryMessageHeaderModel,
+  ): Promise<HistoryMessageHeaderModel | undefined> {
+    await this.delete(id)
+    return await this.messageRepository.save(data)
+  }
+
   public async clean(): Promise<HistoryMessageHeaderModel[] | undefined> {
     const query: HistoryMessageHeaderEntity[] | undefined = await this.messageRepository.find()
     if (!query) {
--- a/src/utils/mqttUtils.ts
+++ b/src/utils/mqttUtils.ts
@@ -143,16 +143,18 @@ export const hasMessagePayload = async (data: HistoryMessagePayloadModel): Promi
   return false
 }
 
-export const hasMessageHeader = async (data: HistoryMessageHeaderModel): Promise<boolean> => {
+export const hasMessageHeader = async (
+  data: HistoryMessageHeaderModel,
+): Promise<HistoryMessageHeaderModel | undefined> => {
   const { historyMessageHeaderService } = useServices()
   const headers = await historyMessageHeaderService.getAll()
   if (headers) {
-    const res = headers.some((el: HistoryMessageHeaderModel) => {
+    const res = headers.find((el: HistoryMessageHeaderModel) => {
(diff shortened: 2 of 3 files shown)


src/database/services/HistoryMessageHeaderService.ts
```diff
@@ -38,6 +38,14 @@ export default class HistoryMessageHeaderService {
     return await this.messageRepository.save(data)
   }
 
+  public async orderChange(
```
oceanlvr, line 41: yes, if we insert message like the order below:

1. topic `test/1`
2. topic `test/2`
3. topic `test/1`

We still get the older one, which is `test/2`, nor did we want.

But I don't think it is a good way, just remove and insert it to sqlite.

Notice we got a field name `updateAt`. Maybe we should fetch data sort by `updateAt`, and we insert a existing message payload, we should just update the time.

Besides, `orderChange` may change a better name, it's a little confusing.

oceanlvr: And if so, I think it same for the message payload, maybe I'll do some check.

ysfscream:
@lileiemq So the real reason why the problem occurred to understand? Is it because the historical data is not updated? Or is it because of the selection of historical data? @oceanlvr Please help check the completion of the reason check, which can be discussed in this PR, before understanding the reason, using the delete and then add is really not a good method.

src/database/services/HistoryMessageHeaderService.ts
lileiemq: It should be that the historical data has not been updated. I am not familiar with sqlite at present.

oceanlvr:
It caused by incorrect data comparison algorithm.

1. send topic `test/1`, it remains a record `test/1` to the history array.   historyMessage = [`test/1`]
2. send topic `test/2`, it push `test/2` to history array. historyMessage = [`test/1`,`test/2`]
3. send topic `test/1`, we comparison `test/1` to history record, and find it has exist then do nothing which is wrong.

We should update the update service API and the data comparison algorithm.

Not just operation with insert and delete the message entity.

oceanlvr:
@lileiemq 
Maybe we need to update current getAll service API

HistoryMessageHeaderService.ts

getAll sort the message model by updateAt filed.

```diff
  public async getAll(): Promise<HistoryMessageHeaderModel[] | undefined> {
-  const query: HistoryMessageHeaderEntity[] | undefined = await this.messageRepository.find()
+  const query: HistoryMessageHeaderEntity[] | undefined = await this.messageRepository
+  .createQueryBuilder('msg')
+  .addOrderBy('updateAt','ASC')
+ .getMany()
    if (!query) {
      return
    }
    return query
  }
```


```diff
export const hasMessageHeader = async (data: HistoryMessageHeaderModel): Promise<boolean> => {
  const { historyMessageHeaderService } = useServices()
  const headers = await historyMessageHeaderService.getAll()
  if (headers) {  
    const res = headers.some((el: HistoryMessageHeaderModel) => {
      return data.qos === el.qos && data.topic === el.topic && data.retain === el.retain
    })
+ // TODO: to update this header's update time
    return res
  }
  return false
}
```

src/database/services/HistoryMessageHeaderService.ts
oceanlvr: We can discuss the historical service API together.

PR #978, opened 2022-07-06
feat(src/main/updateChecker.ts): add a feature for automatic updates

laybrid:
Test the automatic update feature

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "MQTTX",
-  "version": "1.8.0",
+  "version": "1.7.3",
   "description": "MQTT desktop client",
   "author": "EMQX Team <yusf@emqx.io>",
   "scripts": {
@@ -36,6 +36,8 @@
     "chart.js": "^2.9.4",
     "core-js": "^2.6.11",
     "csvtojson": "^2.0.10",
+    "electron-store": "^8.0.2",
+    "electron-updater": "^5.0.5",
     "element-ui": "^2.15.5",
     "fs-extra": "^8.1.0",
     "json2csv": "^5.0.3",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "MQTTX",
-  "version": "1.7.3",
+  "version": "1.8.0",
   "description": "MQTT desktop client",
   "author": "EMQX Team <yusf@emqx.io>",
   "scripts": {
--- a/src/background.ts
+++ b/src/background.ts
@@ -4,7 +4,7 @@ import { app, protocol, BrowserWindow, ipcMain, shell, Menu } from 'electron'
 import { createProtocol } from 'vue-cli-plugin-electron-builder/lib'
 import installExtension, { VUEJS_DEVTOOLS } from 'electron-devtools-installer'
 import { quitAndRenameLogger } from './utils/logger'
-import updateChecker from './main/updateChecker'
+import updateChecker, { createUpdateWindow } from './main/updateChecker'
 import getMenuTemplate from './main/getMenuTemplate'
 import saveFile from './main/saveFile'
 import saveExcel from './main/saveExcel'
@@ -15,6 +15,8 @@ import useServices from '@/database/useServices'
 
 declare const __static: string
 
+const Store = require('electron-store')
+const electronStore = new Store()
 let theme: Theme = 'light'
 let syncOsTheme = false
 let autoCheckUpdate: boolean = true
@@ -149,6 +151,8 @@ async function createWindow() {
   if (autoCheckUpdate) {
     updateChecker()
   }
+  // updateWindow
+  electronStore.get('isShow') && createUpdateWindow()
 }
 
 // This method will be called when Electron has finished
--- a/src/main/updateChecker.ts
+++ b/src/main/updateChecker.ts
@@ -25,11 +25,9 @@ const isUpdate = (latest: string, current: string): boolean => {
 }
 
 const autoDownload = (latest: string, language: string): void => {
-  if (language === 'zh') {
-    autoUpdater.setFeedURL(`https://www.emqx.com/zh/downloads/MQTTX/${latest}`)
-  } else {
-    autoUpdater.setFeedURL(`https://www.emqx.com/en/downloads/MQTTX/${latest}`)
-  }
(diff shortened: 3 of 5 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "MQTTX",
-  "version": "1.8.0",
+  "version": "1.7.3",
```
ysfscream, line 3: Please change the current version to 1.8.0

src/background.ts
```diff
@@ -15,6 +15,8 @@ import useServices from '@/database/useServices'
 
 declare const __static: string
 
+const Store = require('electron-store')
```
ysfscream, line 18: Is it not possible to use the es module?

```js
import Store from 'electron-store'
```

src/main/updateChecker.ts
```diff
@@ -1,9 +1,14 @@
-import { dialog, shell } from 'electron'
+import { dialog, BrowserWindow } from 'electron'
 import axios from 'axios'
 import version from '@/version'
+import useServices from '@/database/useServices'
+
+const { autoUpdater } = require('electron-updater')
+const Store = require('electron-store')
```
ysfscream, line 7: Same here, Why no ES module?

src/main/updateChecker.ts
```diff
@@ -19,25 +24,54 @@ const isUpdate = (latest: string, current: string): boolean => {
...
+  autoUpdater.on('error', () => {})
+  autoUpdater.on('download-progress', () => {})
+  autoUpdater.on('update-downloaded', () => {
+    electronStore.set('isShow', true)
+    dialog
+      .showMessageBox({
+        type: 'info',
+        title: 'New Version',
+        buttons: ['Download', 'No'],
+        message: `Update available:${latest}版本`,
```
ysfscream, line 46:
```suggestion
        message: `Update available: ${latest}`,
```

ysfscream, line 46: No Chinese:

```suggestion
        message: `Update available: ${latest}`,
```

vue.config.js
```diff
@@ -7,6 +7,9 @@ module.exports = {
     host: '0.0.0.0',
     disableHostCheck: true,
   },
+  css: {
+    extract: false,
+  },
```
ysfscream, line 12: Why is extract set to false?

src/background.ts
laybrid: Of course, but it is more convenient to implement the function with this.

ysfscream: However, for the whole project, it is better to use the ES Module approach to import if it is not necessary and needs to be unified

src/main/updateChecker.ts
laybrid: Already answered

src/main/updateChecker.ts
```diff
@@ -19,25 +24,54 @@ const isUpdate = (latest: string, current: string): boolean => {
   return update
 }
 
+const autoDownload = (latest: string, language: string): void => {
+  if (language === 'zh') {
+    autoUpdater.setFeedURL(`https://www.emqx.com/zh/downloads/MQTTX/v1.8.0`)
+  } else {
+    autoUpdater.setFeedURL(`https://www.emqx.com/en/downloads/MQTTX/v1.8.0`)
+  }
```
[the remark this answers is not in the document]
laybrid: Will be changed to
` autoUpdater.setFeedURL(`https://www.emqx.com/zh/downloads/MQTTX/${latest}`)`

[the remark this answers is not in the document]
ysfscream: The latest version is available here: https://github.com/emqx/MQTTX/blob/main/src/version.ts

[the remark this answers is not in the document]
laybrid: Will be changed to
`
 autoUpdater.setFeedURL(`https://www.emqx.com/zh/downloads/MQTTX/${latest}`)
`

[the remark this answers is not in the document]
laybrid: It will be changed to concatenated string.

src/main/updateChecker.ts
laybrid: This will change later.

vue.config.js
laybrid: Extract CSS is not turned on by default in development environment mode because it is incompatible with CSS hot overloading.

src/background.ts
laybrid: Sorry, I just misunderstood. You can't use es Module here.

package.json
oceanlvr: maybe there should be based on 1.8.0 version

src/background.ts
oceanlvr: could we use typeorm insteadof electron-store?

oceanlvr: `electron-store` is a CJS module, and [vue-cli-plugin-electron-builder](https://github.com/nklayman/vue-cli-plugin-electron-builder) help us transform `ESM import` to CJS require.
So yes you are actually using CJS module, but the plugin does the conversion for you(ESM import to cjs require), please rest assured to use ESM import.

oceanlvr: I think typeorm and electron-store should not be used together, you should always ensure that there is only one state disk persistence library

src/main/updateChecker.ts
```diff
@@ -19,25 +24,54 @@ const isUpdate = (latest: string, current: string): boolean => {
...
+        message: `Update available:${latest}版本`,
+      })
+      .then((res) => {
+        if (res.response === 0) {
+          // if selected yes
+          autoUpdater.quitAndInstall()
+        } else {
+          dialog.showMessageBox({
+            type: 'info',
+            message: 'Automatic update on do not shut down the computer immediately',
```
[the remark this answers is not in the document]
oceanlvr: I see, It turns out that i18n can be transformed into a public module in the future, but there are fewer hints that the main process needs i18n.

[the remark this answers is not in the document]
ysfscream: Yes, but it should not be discussed in this PR, perhaps we should have some optimization in the next.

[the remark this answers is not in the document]
oceanlvr: Got it, we just ignore it this version~

PR #1142, opened 2022-11-14
feat(desktop & web): support auto scroll interval setting

Red-Asuka:
### PR Checklist

#### Issue Number

Example: \#123

Support auto scroll interval setting.

image

- [ ] Yes
- [ ] No

#### Specific Instructions


src/views/connections/ConnectionsDetail.vue
```diff
@@ -1457,6 +1465,12 @@ export default class ConnectionsDetail extends Vue {
 
   private created() {
     this.getConnectionValue(this.curConnectionId)
+    this.scrollSubject
+      .asObservable()
+      .pipe(throttleTime(this.autoScrollInterval * 1000))
+      .subscribe(() => {
+        this.scrollToBottom()
+      })
```
ysfscream, line 1473: Should we think about the users who enabled the `autoScroll`?

Because of this, we need not set this `Subject`?

Red-Asuka: The judgment is inside the execution function, for performance in the external judgment again better, but the code logic is duplicated, what do you think?

```ts
  private scrollToBottom() {
    if (this.autoScroll === false) {
      return
    }
    const timer = setTimeout(() => {
      const messagesDisplay = this.$refs.messagesDisplay as MessageList
      const messagesDisplayDOM = messagesDisplay.$el
      if (messagesDisplayDOM) {
        messagesDisplayDOM.scrollTo({
          top: messagesDisplayDOM.scrollHeight + 160,
          left: 0,
          behavior: 'smooth',
        })
      }
      clearTimeout(timer)
    }, 100)
  }
```

ysfscream: > or performance in the external judgment again better

I agree with this.

I don't think the logic of the code will be duplicated, because just outside the function to determine the condition is good, and inside we just have to execute the scrolling function. It's right?

Red-Asuka: You are right, I will determine the conditions before subscribe.

Red-Asuka: https://github.com/emqx/MQTTX/pull/1142/commits/486466f03e61a76aefe314efe1fc549b448b45db

ysfscream: > private scrollToBottom() {
    if (this.autoScroll === false) {
      return
    }
    const timer = setTimeout(() => {
      const messagesDisplay = this.$refs.messagesDisplay as MessageList
      const messagesDisplayDOM = messagesDisplay.$el
      ...

And should we remove the `autoScroll` condition here? Not needed anymore?

Red-Asuka: The conditions here can't be removed, because some behavior-triggered scrolling don't throttle through the `Subject`, but directly called scrolling method, which are prepared for such behaviors.

ysfscream: You mean something like the user active the topic or payload input?

Red-Asuka: Yes

ysfscream (approved):

PR #1157, opened 2022-11-30
fix(cli): allow multiple occurrences of the same name in user properties

Red-Asuka:
### PR Checklist

#### Issue Number

Example: \#123

image

- [ ] Yes
- [x] No

#### Specific Instructions

### DIFF
--- a/cli/src/utils/parse.ts
+++ b/cli/src/utils/parse.ts
@@ -34,10 +34,23 @@ const parseMQTTVersion = (value: string) => {
   return dict[value as '3.1' | '3.1.1' | '5']
 }
 
-const parseUserProperties = (value: string, previous: Record<string, unknown> | undefined) => {
+const parseUserProperties = (value: string, previous?: Record<string, string | string[]>) => {
   const [key, val] = value.split(': ')
   if (key && val) {
-    return previous ? { ...previous, [key]: val } : { [key]: val }
+    if (!previous) {
+      return { [key]: val }
+    } else {
+      if (previous[key]) {
+        if (Array.isArray(previous[key])) {
+          ;(previous[key] as string[]).push(val)
+        } else {
+          previous[key] = [previous[key] as string, val]
+        }
+        return previous
+      } else {
+        return { ...previous, [key]: val }
+      }
+    }
   } else {
     signale.error('Not a valid user properties.')
     process.exit(1)
--- a/cli/src/lib/sub.ts
+++ b/cli/src/lib/sub.ts
@@ -67,8 +67,19 @@ const sub = (options: SubscribeOptions) => {
 
     packet.retain && msgData.push({ label: 'retain', value: packet.retain })
 
-    packet.properties?.userProperties &&
-      msgData.push({ label: 'user properties', value: { ...packet.properties.userProperties } })
+    if (packet.properties?.userProperties) {
+      const up: { key: string; value: string }[] = []
+      Object.entries(packet.properties.userProperties).forEach(([key, value]) => {
+        if (typeof value === 'string') {
+          up.push({ key, value })
+        } else {
+          value.forEach((v) => {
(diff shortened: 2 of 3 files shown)


cli/src/utils/parse.ts
```diff
@@ -34,10 +34,23 @@ const parseMQTTVersion = (value: string) => {
...
 }
 
-const parseUserProperties = (value: string, previous: Record<string, unknown> | undefined) => {
+const parseUserProperties = (value: string, previous?: Record<string, string | string[]>) => {
   const [key, val] = value.split(': ')
   if (key && val) {
-    return previous ? { ...previous, [key]: val } : { [key]: val }
+    if (!previous) {
+      return { [key]: val }
+    } else {
```
ysfscream, line 42: `else` is needless after the `return` statement.

```
if (!previous) return { [key]: value }
if (xxx) {
    xxxx
}

File: cli/src/utils/parse.ts [hunk 2]
```diff
@@ -34,10 +34,23 @@ const parseMQTTVersion = (value: string) => {
...
+      if (previous[key]) {
+        if (Array.isArray(previous[key])) {
+          ;(previous[key] as string[]).push(val)
```
ysfscream, line 45: Please declare an internal variable to represent `previous[key]`, For example:

```js
const objValues = previous[key] as string[]
```

Looks cumbersome, but more clear type

cli/src/utils/parse.ts
Red-Asuka: I know that `else` is unnecessary, but I think that in some cases `else` can make the code clearer. This is the code without `else`, what do you think?

```ts
    if (!previous) return { [key]: val }
    if (previous[key]) {
      if (Array.isArray(previous[key])) {
        ;(previous[key] as string[]).push(val)
      } else {
        previous[key] = [previous[key] as string, val]
      }
      return previous
    } else {
      return { ...previous, [key]: val }
    }
```

Red-Asuka: The final code:

```ts
    if (!previous) return { [key]: val }
    if (previous[key]) {
      if (Array.isArray(previous[key])) {
        const currentVal = previous[key] as string[]
        currentVal.push(val)
        return { ...previous, [key]: currentVal }
      }
      previous[key] = [previous[key] as string, val]
      return previous
    }
    return { ...previous, [key]: val }
```

Red-Asuka: Or
```ts
    if (!previous) return { [key]: val }
    if (!previous[key]) return { ...previous, [key]: val }
    if (Array.isArray(previous[key])) {
      const currentVal = previous[key] as string[]
      currentVal.push(val)
      return { ...previous, [key]: currentVal }
    }
    previous[key] = [previous[key] as string, val]
    return previous
```

ysfscream: `else` needs continuity, and `return` means execution is over. It doesn't say here that else doesn't need to be used, I think this code is just fine

```js
if (!previous) return { [key]: val }
if (previous[key]) {
  if (Array.isArray(previous[key])) {
	const currentVal = previous[key] as string[]
     currentVal.push(val)
  } else {
  	const currentVal = previous[key] as string
    previous[key] = [currentVal, val]
  }
  return previous
}
 return { ...previous, [key]: val }
```

ysfscream (approved):

cli/src/utils/parse.ts
ysfscream: Anyway, I will merge first. But the principle of my judgment is that if there is a return in the if, there will be no need for else; if not, then there is a need for else; this proposal should be in the rules of `Eslint` also had, personal preference, code readability is our goal.

ysfscream: It's a nice job.

Issue #1104, opened 2022-10-12
[Bug] It's completely stuck after long period running.


brunofgc:
#### What happened?

After updated to the last version (1.8.3) I could see it's good to use for "long periods".

It was working fine for hours something like 4 hours without getting freeze.

But now, when I try to open MQTTX again, it goes above 3.1GB of RAM, and seems to be retrieving some "cached data". But it stops working and just get a black screen instead of the program GUI. (The GUI disapears and then I get the black screen). (I'm using "night mode").

I tried to unstall and install again in order to "delete the cached data" but it didn't work.

It's a very good program! The best in my opinion and the best fit for my application. I hope it will be better day by day.

Thank you for providing such a good tool!

#### Environment

- OS: Windows 11 64Bit (Intel Core I7)
- MQTTX version, e.g. v1.8.3

#### More detail

![image](https://user-images.githubusercontent.com/2005246/195387883-0d0e680d-820c-4690-a43d-8e34aa13d601.png)

ysfscream:
Thank you for your use and sorry for the trouble, or maybe this is the duplicate of #968

The `MQTTX.db` file needs to be deleted, but my question is this, are you receiving a lot of messages all the time while you are running?

brunofgc:
Thanks for answering. Yes! It is receiving a lot of messages all the time.

Em qui, 13 de out de 2022 00:22, YuShifan ***@***.***>
escreveu:

ysfscream:
Yes, there is always an error in the MQTT X database when saving a large number of messages, this is a known issue, especially on Windows. Next, we will continue to look at how to optimize, thank you very much for your support.

LucidEye:
2 months later and still no resolution on this bug????  Seriously, what did you guys expect would happen when you start subscribing to multiple topics from multiple publishers, and also cache topic messages while the app is shut down?!?!?!?!  How could you possibly overlook the flood of data entries that would cause every time you open the app?!?!?!?  Pardon my French, but that is just piss poor development work.  Maybe you should try a different database that can handle the load of so many time series entries at once (and maybe a better DBA to do the coding)?

ysfscream:
Apologies for the long wait, welcome to https://github.com/emqx/MQTTX/releases/tag/v1.9.1, that should solve the problem and I will close this issue.

PR #1191, opened 2023-01-11
improve(desktop): refactor the message rendering logic

Red-Asuka:

The messages are loaded in full, resulting in increased memory usage as the number of messages increases and causing the page to become slow. When the number of messages becomes too large, it can make the application unusable.

#### Issue Number

\#1104

Memory usage will be kept within 200MB and will not increase as the number of messages increases. The MQTT X will still be able to function smoothly when faced with a large number of messages.

- Refactor the message rendering logic to use a dynamic infinite scrolling list
- Change message filtering and topic filtering to use SQL queries
- Remove the automatic scrolling setting
- Add new message notifications

image

- [x] Yes
- [ ] No

#### Specific Instructions


src/database/services/MessageService.ts
```diff
@@ -57,6 +57,21 @@ export default class MessageService {
     }
   }
 
+  public async getMore(connectionId: string, createAt: string, mode: 'before' | 'after' = 'before', limit = 20) {
```
ysfscream, line 60: loadMore

src/lang/common.ts
```diff
@@ -258,4 +258,18 @@ export default {
     ja: '期限なし',
     hu: 'Soha ne járjon le',
   },
+  newMsg: {
+    zh: '条新消息',
+    en: 'new message',
```
ysfscream, line 263: {0} new message | messages

src/lang/common.ts
```diff
@@ -258,4 +258,18 @@ export default {
...
+    ja: '新しいメッセージ',
+    hu: 'új üzenet',
+  },
+  newMsgs: {
+    zh: '条新消息',
+    en: 'new messages',
+    tr: 'yeni mesajlar',
+    ja: '新しいメッセージ',
+    hu: 'új üzenetek',
+  },
```
ysfscream, line 274: remove it

src/lang/common.ts
ysfscream: https://kazupon.github.io/vue-i18n/guide/pluralization.html#accessing-the-number-via-the-pre-defined-argument

ysfscream (approved):

ysfscream:
I will merge it first for testing.

PR #1269, opened 2023-04-24
feat(simulation): add simulator to the bench

wivwiv:

Users can place a custom script in the simulator directory and then use it to generate simulation data in the bench pub with the -sim option.

This helps you to achieve a simulated performance test that is closer to the production.

```bash
mqttx bench pub -sim tesla -t t/1 -c 1

mqttx sub -t t/1
{"car_id":"JK6ZSF1A59HS94294","display_name":"Ines's Tesla","model":"Y","trim_badging":"ullam","exterior_color":"purple","wheel_type":"repudiandae","spoiler_type":"placeat","geofence":"Longview","state":"charging","since":"2023-04-23T18:25:00.558Z","healthy":true,"version":"0.2.9","update_available":false,"update_version":"1.0.2","latitude":"-15.3417","longitude":"127.2175","shift_state":"P","power":-6046,"speed":189,"heading":50,"elevation":3851,"locked":true,"sentry_mode":true,"windows_open":false,"doors_open":true,"trunk_open":true,"frunk_open":false,"is_user_present":false,"is_climate_on":false,"inside_temp":38.4,"outside_temp":44.8,"is_preconditioning":true,"odometer":383942,"est_battery_range_km":653.7,"rated_battery_range_km":916.6,"ideal_battery_range_km":542.9,"battery_level":4,"usable_battery_level":87,"plugged_in":true,"charge_energy_added":74.07,"charge_limit_soc":42,"charge_port_door_open":true,"charger_actual_current":47.44,"charger_power":47,"charger_voltage":240,"charge_current_request":15,"charge_current_request_max":20,"scheduled_charging_start_time":"2023-01-25T02:10:33.559Z","time_to_full_charge":8.42,"tpms_pressure_fl":2.5,"tpms_pressure_fr":3.3,"tpms_pressure_rl":3.2,"tpms_pressure_rr":2.3,"timestamp":1682330389155}
```

- [ ] Yes
- [x] No

#### Specific Instructions

### DIFF
--- a/cli/src/index.ts
+++ b/cli/src/index.ts
@@ -12,6 +12,7 @@ import {
   parsePubTopic,
   parseFormat,
   parseOutputMode,
+  parseSimulator,
 } from './utils/parse'
 import { conn, benchConn } from './lib/conn'
 import { pub, benchPub } from './lib/pub'
@@ -355,6 +356,7 @@ export class Commander {
         'the message topic, support %u (username), %c (client id), %i (index) variables',
         parsePubTopic,
       )
+      .option('-sim, --simulator <SIMULATOR>', 'message generation simulator', parseSimulator, '')
       .option('-m, --message <BODY>', 'the message body', 'Hello From MQTT X CLI')
       .option('-q, --qos <0/1/2>', 'the QoS of the message', parseNumber, 0)
       .option('-r, --retain', 'send a retained message')
--- a/cli/README-CN.md
+++ b/cli/README-CN.md
@@ -355,6 +355,8 @@ mqttx bench pub --help
 | -im, --message-interval <MILLISECONDS> | 发布消息的间隔时间，单位为毫秒，默认为 1000ms                        |
 | -I, --client-id <ID>                   | 客户端 ID，支持 %i (索引) 占位符                                     |
 | -t, --topic <TOPIC...>                 | 需要订阅的 Topic, 支持 %u (用户名), %c (客户端 ID), %i (索引) 占位符 |
+| -m, --message <MSG>                               | 需要发布的 Payload 消息                                      |
+| -sim, --simulator <SIMULATOR>           | 使用模拟器生成特定场景的消息，会覆盖 -m 参数 |
 | -v, --verbose                          | 打印发送出的历史消息数量与消息速率                                   |
 | ~~-s, --stdin~~                        | ~~从 stdin 中读取信息体~~                                            |
 | ~~-M, --multiline~~                    | ~~可以通过多行发布多条消息~~                                         |
--- a/cli/README.md
+++ b/cli/README.md
@@ -354,6 +354,8 @@ mqttx bench pub --help
 | -im, --interval-message <MILLISECONDS> | interval of publishing message to the broker (default: 1000ms)                 |
 | -I, --client-id <ID>                   | the client id, support %i (index) variable                                     |
 | -t, --topic <TOPIC...>                 | the message topic, support %u (username), %c (client id), %i (index) variables |
+| -m, --message <MSG>                               | the message body (default: "Hello From MQTT X CLI")                                                                                |
+| -sim, --simulator <SIMULATOR> | Using the simulator to generate scenario-specific messages, it will override the -m parameter |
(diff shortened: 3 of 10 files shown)


Red-Asuka:
@wivwiv How to introduce an external custom simulator? I see that the directory in the code is obtained using the following method, which is the path of the current file. The simulator needs to be in this directory to work properly. The development environment directory address is `MQTTX/cli/dist/src/simulator`. When users use it, it is packaged into a binary executable file. I am not sure how to reference it specifically.

```ts
const SimulatorFolder = path.join(__dirname, '../simulator')
```

Red-Asuka:
@wivwiv CLI currently does not have a configuration to automatically format code, nor does it use ESLint. It only has a Prettier configuration file. Please use Prettier to format the code before submitting.

cli/src/index.ts
```diff
@@ -355,6 +356,7 @@ export class Commander {
         'the message topic, support %u (username), %c (client id), %i (index) variables',
         parsePubTopic,
       )
+      .option('-sim, --simulator <SIMULATOR>', 'message generation simulator', parseSimulator, '')
```
Red-Asuka, line 359: Maybe this feature can also be added to the `pub` command

ysfscream:
Thank you for submitting this PR! After reviewing the changes, I have a few suggestions to make the mqttx cli library more flexible and adaptable for different use cases.

1. Remove the 'tesla' data file: As discussed, the 'tesla' file is quite specific to a particular use case. Including it in the library may not be suitable for a general-purpose utility like mqttx cli. I recommend removing this file from the PR.

2. Implement pluggable scenario templates: Instead of including a specific data file, consider introducing a pluggable approach that allows users to create their scenario data templates based on their needs. This would make the library more versatile and cater to a wider range of use cases.

3. Provide examples: To guide users in creating their templates, including a few example templates showcasing different use cases would be helpful. This will assist users in understanding how to use the library effectively and create a more user-friendly experience.

Once these changes are implemented, I believe the PR will be a valuable addition to the mqttx cli library. Thank you for your hard work EMQX PM 😂, and please let me know if you have any questions or concerns!

ysfscream:
1. **Define a template function**: Create a template function that takes the scenario configuration as input and returns an object containing the topic and payload.

```javascript
function defaultTemplate(config) {
    return {
        topic: 'default_topic',
        payload: 'default_payload'
    };
}
```

2. **Allow users to pass a custom template**: Update your `mqttx` CLI tool to accept a custom template function as an optional parameter. If no template is provided, the default implementation will be used.

```javascript
// mqttx-bench.js
const yargs = require('yargs/yargs');
const { hideBin } = require('yargs/helpers');

// Import the defaultTemplate function
const { defaultTemplate } = require('./template');

const argv = yargs(hideBin(process.argv))
    .command('bench pub [config] [template]', 'publish benchmark', (yargs) => {
        yargs
            .positional('config', {
                describe: 'path to the configuration file',
                type: 'string'
            })
            .positional('template', {
                describe: 'path to the custom template function file',
                type: 'string'
            });
    })
    .help()
    .argv;

function runMqttxCli(configPath, templatePath) {
    const config = require(configPath);
    const template = templatePath ? require(templatePath) : defaultTemplate;

    // Use the provided template function to get topic and payload
    const { topic, payload } = template(config);

    // Continue with the rest of the mqttx CLI logic
}
```

3. **Update documentation**: Update your library's documentation to explain how users can create their custom templates as functions and pass the path of the function file as an argument to the `mqttx bench pub` command.

Here's an example of a custom template function implementation in a separate file, let's say `my_custom_template.js`:

```javascript
// my_custom_template.js
function myCustomTemplate(config) {
    return {
        topic: 'my_custom_topic',
        payload: 'my_custom_payload'
    };
}

module.exports = myCustomTemplate;
```

Users can then run the MQTTX CLI tool with their custom template like this:

```bash
mqttx bench pub ./path/to/config.json ./path/to/my_custom_template.js
```

By implementing pluggable scenario templates using functions, you allow users to customize the behavior of your MQTTX CLI tool for different use cases.

ysfscream (changes requested):
Need Update

PR #1280, opened 2023-05-06
feat(simulate): add simulate to cli

wivwiv:

Users can use the new `simulate` command to simulate publishing messages in specific scenarios, such as the status reported by hundreds of telsa model 3 per second.

It helps users to generate a large number of the most realistic dynamic messages in a short time.

Although it sounds like a subset of bench, it should be a completely new feature.  

In the future it will be built with rich scenarios for IIoT, IoV, Energy and Power, etc. It also allows users to customize the simulator —— specify a `js` file to perform data generation, or even replay historical data from a database or stream.

**Usage**

```
$ mqttx simulate -sc tesla -c 10
[2023/5/6] [12:19:08] › ℹ  Start simulation publishing, scenario: tesla, connections: 10, req interval: 10ms, message interval: 1000m

$ mqttx simulate -f  ./dist/src/scenarios/tesla.js -c 10
2023/5/6] [12:21:53] › ℹ  Start simulation publishing, scenario: ./dist/src/scenarios/tesla.js, connections: 10, req interval: 10ms, message interval: 1000ms
```

```
$ mqttx sub -t +/# -v
[2023-5-6] [12:20:00] › topic: mqttx/simulate/tesla/mqttx_0038ee3e
payload: {"car_id":"XSDW6ET1XLDJ17357","display_name":"Chloe's Tesla","model":"S","trim_badging":"excepturi","exterior_color":"lavender","wheel_type":"beatae","spoiler_type":"autem","geofence":"Port Cletaburgh","state":"online","since":"2023-05-05T14:51:44.814Z","healthy":false,"version":"6.6.1", ...}
```

- [ ] Yes
- [x] No

#### Specific Instructions

### DIFF
--- a/cli/README.md
+++ b/cli/README.md
@@ -162,7 +162,7 @@ mqttx --help
 | pub     | Publish a message to a topic                   |
 | sub     | Subscribes to one or multiple topics           |
 | bench   | MQTT Benchmark in performance testing          |
-| simulate \| sim | Simulate publishing scenario-specific MQTT messages |
+| simulate | Simulate publishing scenario-specific MQTT messages |
 
 #### Connect
 
@@ -376,7 +376,7 @@ For simulating MQTT publish message in specific scenarios.
 It has basically the same as the [Publish Benchmark](#publish-benchmark) command options, the following will only list the new or changed options.
 
 ```shell
-mqttx sim --help
+mqttx simulate --help
 ```
 
 | Options                          | Description                             |
@@ -386,6 +386,28 @@ mqttx sim --help
 | -t, --topic <TOPIC... > | the message topic, optional, support %u (username), %c (client id), %i (index),  %sc (scenario) variables, defaults to `mqttx/simulate/%sc/%c` |
 
 One of the `--scenario` and `--file` parameters must be specified, and if both are specified, the `--file` parameter is preferred.
+
+Scenario file example:
+
+```js
+function generator (option, clientId) {
+  return {
+    // topic: 'mqttx/simulate/myScenario/' + clientId,
+    message: JSON.stringify({
+      temp: Math.random() * 10 + 20,
+      hum: Math.random() * 10 + 60,
+    })
+  }
+}
+module.exports = {
+  name: 'myScenario',
+  generator,
+  // author: 'mqttx',
+  // version: '1.0.0',
+  // description: 'my scenario description',
+  // dataFormat: 'JSON'
+}
+```
 
 ## Better Together with EMQX
 
--- a/cli/README.md
+++ b/cli/README.md
@@ -392,6 +392,7 @@ Scenario file example:
 ```js
 function generator (option, clientId) {
   return {
+    // If no topic is returned, use the topic in the command line parameters
     // topic: 'mqttx/simulate/myScenario/' + clientId,
     message: JSON.stringify({
       temp: Math.random() * 10 + 20,
@@ -402,10 +403,6 @@ function generator (option, clientId) {
 module.exports = {
   name: 'myScenario',
   generator,
-  // author: 'mqttx',
-  // version: '1.0.0',
-  // description: 'my scenario description',
-  // dataFormat: 'JSON'
 }
 ```
 
--- a/cli/README.md
+++ b/cli/README.md
@@ -134,6 +134,16 @@ mqttx bench sub -c 5000 -t bench/%i
 mqttx bench pub -c 5000 -t bench/%i
 ```
 
+Simulate
+
+```bash
+# Specify a local scenario and start the simulation
+mqttx simulate -sc tesla -c 10
+
+# Specify a scenario file and start the simulation
+mqttx simulate -f <scenario file path> -c 10
+```
+
 ### Help
 
 ```shell
@@ -152,6 +162,7 @@ mqttx --help
 | pub     | Publish a message to a topic                   |
 | sub     | Subscribes to one or multiple topics           |
 | bench   | MQTT Benchmark in performance testing          |
+| simulate \| sim | Simulate publishing scenario-specific MQTT messages |
 
 #### Connect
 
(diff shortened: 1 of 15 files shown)


cli/src/types/global.d.ts
```diff
@@ -106,6 +106,15 @@ declare global {
     interval: number
   }
 
+  interface SimulatePubOptions extends BenchPublishOptions {
+    scenario: string
+    count: number
+    interval: number
+    messageInterval: number
+    verbose: boolean
```
Red-Asuka, line 114: These items are already defined in `BenchPublishOptions`.

```
count: number
interval: number
messageInterval: number
verbose: boolean
```

cli/README.md
```diff
@@ -358,6 +369,24 @@ mqttx bench pub --help
...
+mqttx sim --help
+```
+
+| Options                          | Description                             |
+| ----------------------------- | -------------------------------- |
+| -sc, --scenario <SCENARIO>          | the name of the local scenario to simulate           |
+| -f, --file <SCENARIO FILE PATH> | the path of the scenario file  |
+| -t, --topic <TOPIC... > | the message topic, optional, support %u (username), %c (client id), %i (index),  %sc (scenario) variables, defaults to `mqttx/simulate/%sc/%c` |
+
+One of the `--scenario` and `--file` parameters must be specified, and if both are specified, the `--file` parameter is preferred.
```
Red-Asuka, line 388: I see that the simulator requires a generator method to return data and defaults to passing some parameters. An appropriate explanation should be provided here, along with a link to the simplest example of a simulator.

cli/src/lib/pub.ts
```diff
@@ -178,7 +185,19 @@ const benchPub = async (options: BenchPublishOptions) => {
           interactive.success('[%d/%d] - Connected', connectedCount, count)
 
           setInterval(() => {
-            client.connected &&
+            if (simulator) {
+              const { topic, message } = simulator.generator(options as SimulatePubOptions, client.options.clientId)
+              client.connected &&
+              client.publish(topic || topicName, message, pubOpts.opts, (err) => {
```
Red-Asuka, line 191: The processing method for "publish" has not changed, only the input parameters are different. Please optimize the code.

cli/src/utils/simulate.ts
```diff
@@ -0,0 +1,68 @@
...
+
+const getScenarioFilePath = function (file: string): string {
+  const absoluePath = path.resolve(file)
+  if (fs.existsSync(absoluePath)) {
+    return absoluePath
+  }
+  return ''
+}
+
+interface ISimulator {
```
Red-Asuka, line 27: Move to `global.d.ts`.

cli/tsconfig.json
```diff
@@ -10,5 +11,5 @@
     "resolveJsonModule": true,
     "types": ["@types/signale"]
   },
-  "include": ["src"]
+  "include": ["src", "scenarios/*.js"]
```
Red-Asuka, line 14: Directly write `scenarios` to `/cli/src/scenarios/`, and use TypeScript.

cli/tsconfig.json
```diff
@@ -1,5 +1,6 @@
 {
   "compilerOptions": {
+    "allowJs": true,
```
Red-Asuka, line 3: Delete it.

scripts/scenarios/tesla.js
```diff
@@ -0,0 +1,87 @@
+const faker = require('faker')
```
Red-Asuka, line 1: Directly write `scenarios` to `/cli/src/scenarios/`, and use TypeScript.

Red-Asuka (changes requested):

Red-Asuka:
Please use Prettier to format the code.

image

Red-Asuka:
Please provide users with the simplest example of a simulator and explain it in the documentation.

cli/README.md
```diff
@@ -358,6 +369,24 @@ mqttx bench pub --help
...
+
+```shell
+mqttx sim --help
+```
+
+| Options                          | Description                             |
+| ----------------------------- | -------------------------------- |
+| -sc, --scenario <SCENARIO>          | the name of the local scenario to simulate           |
+| -f, --file <SCENARIO FILE PATH> | the path of the scenario file  |
+| -t, --topic <TOPIC... > | the message topic, optional, support %u (username), %c (client id), %i (index),  %sc (scenario) variables, defaults to `mqttx/simulate/%sc/%c` |
```
Red-Asuka, line 386: When using a custom scenario, the default topic %sc is undefined. Can the scenario file name be used instead?

image

cli/README.md
```diff
@@ -152,6 +162,7 @@ mqttx --help
 | pub     | Publish a message to a topic                   |
 | sub     | Subscribes to one or multiple topics           |
 | bench   | MQTT Benchmark in performance testing          |
+| simulate \| sim | Simulate publishing scenario-specific MQTT messages |
```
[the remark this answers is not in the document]
ysfscream: @wivwiv I think no matter `sim` or `simulate`, I think it's better to keep one, and I prefer `sim`, similar to `pub` `sub`.

cli/tsconfig.json
wivwiv: I need to first put `scripts/scenarios/*.js` file to the `cli/src/scenarios` directory to ensure that the script is reusable inside the MQTTX App.

cli/src/types/global.d.ts
```diff
@@ -1,5 +1,5 @@
 declare global {
-  type CommandType = 'conn' | 'pub' | 'sub' | 'benchConn' | 'benchPub' | 'benchSub'
+  type CommandType = 'conn' | 'pub' | 'sub' | 'benchConn' | 'benchPub' | 'benchSub' | 'simulate' | 'sim'
```
ysfscream, line 2: Same here. keep one and I think.

cli/tsconfig.json
wivwiv: For potential users, custom scripts can only use javascript.

- typescript is not widely used
- The `.ts` file may not be executed at runtime in --file mode?

scripts/scenarios/tesla.js
```diff
@@ -0,0 +1,87 @@
...
+const version = '0.0.1'
+const description = 'Simulation to generate Tesla\'s data, reference form https://github.com/adriankumpf/teslamate'
+
+module.exports = {
+  generator,
+  name,
+  dataType,
+  version,
+  description,
+}
```
ysfscream, line 86: Are these unified export fields and data?

cli/tsconfig.json
Red-Asuka: My point is that the built-in scenarios should be written in TypeScript as much as possible, while user-defined ones can only use JavaScript.

ysfscream (commented):
I'm more concerned about what to do if it's a custom local scenario data file. What is the complete process of using custom local scenario data files if not using the built-in simulated datasets?

scripts/scenarios/tesla.js
wivwiv: Yes, in the future it might be possible to add an option to print the simulation data before starting.

This is inspired by the common plugin mechanism, which should always contain meta data for easy exchanges between different users.
A custom script can be treated as a plugin

cli/README.md
wivwiv: fixed

ysfscream (approved):
LGTM

Red-Asuka (approved):

PR #1292, opened 2023-05-17
feat(cli): add ls command

ysfscream:
### PR Checklist

Currently, MQTTX CLI does not allow users to list all the available built-in scenarios. This creates a usability issue since users may not be aware of all the available scenarios they can use.

#### Issue Number

Example: \#123

This PR adds a new ls command to the MQTTX CLI. When used with the --scenarios option, this command will list all the available built-in scenarios and their descriptions. This will make it easier for users to discover and use the built-in scenarios.

image

- [ ] Yes
- [x] No

#### Specific Instructions

### DIFF
--- a/cli/package.json
+++ b/cli/package.json
@@ -20,6 +20,7 @@
   "dependencies": {
     "axios": "^0.27.2",
     "chalk": "~4.1.2",
+    "cli-table3": "^0.6.3",
     "commander": "^9.3.0",
     "compare-versions": "^5.0.1",
     "concat-stream": "^2.0.0",
@@ -53,6 +54,9 @@
       "node16-win-arm64",
       "node16-alpine-arm64"
     ],
-    "outputPath": "release"
+    "outputPath": "release",
+    "scripts": [
+      "dist/src/scenarios/**/*"
+    ]
   }
 }
--- a/cli/src/index.ts
+++ b/cli/src/index.ts
@@ -16,6 +16,7 @@ import {
 import { conn, benchConn } from './lib/conn'
 import { pub, benchPub, simulatePub } from './lib/pub'
 import { sub, benchSub } from './lib/sub'
+import ls from './lib/ls'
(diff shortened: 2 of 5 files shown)


cli/package.json
```diff
@@ -53,6 +54,9 @@
       "node16-win-arm64",
       "node16-alpine-arm64"
     ],
-    "outputPath": "release"
+    "outputPath": "release",
+    "scripts": [
```
Red-Asuka, line 58: "scripts" generally define some running commands, but it seems that they are not used in this way here. What is their purpose here?

ysfscream: "scripts" doesn't refer to npm scripts or similar, but rather a list of file paths that should be bundled into the executable. This is particularly useful for including files that your application needs at runtime, such as dynamic modules or other assets.

ysfscream: If we don't set it when pkg this to binary package, when we run this command will get this error:

```shell
./mqttx-macos-arm64 ls -sc
```

Error:
```shell
Error reading scenarios directory: Error: Directory '/**/cli/dist/scenarios' was not included into executable at compilation stage. Please recompile adding it as asset or script.
```

ysfscream: We can only make this piece of code work properly by setting it up in the pkg configuration:

```js
const dir = await fs.promises.readdir(scenariosDir)
    for (const file of dir) {
      if (path.extname(file) === '.js') {
        // Use dynamic import
        const scenario = await import(path.join(scenariosDir, file))
        scenarios.push({
          name: scenario.name,
          description: scenario.description,
        })
      }
    }
```

This allows us to dynamically import scenario files, because after packaging we are unable to load them.

Red-Asuka: I'm not familiar with this usage. Can you provide more information or point me to any relevant documentation? I'd like to learn more about it.

Red-Asuka (approved):

Issue #916, opened 2022-03-18
[Feature] ALPN protocol support


overheat:
#### Motivation

Many IoT Cloud service provider, like aws, need ALPN protocol for advance usage.
Refer here: https://docs.aws.amazon.com/iot/latest/developerguide/protocols.html

#### Detailed design

A text plain for input ALPN content. 

#### Alternatives

Describe alternatives you've considered, what other designs have been considered? What is the impact of not doing this?

ysfscream:
@overheat Hi, thanks for your advice! We will look at and discuss the suggestions you provide.

muhufuk:
@overheat It would be great If you implement ALPN support to MQTTX

ysfscream:
Thank you for being so patient!

We have addressed and resolved this issue in the https://github.com/emqx/MQTTX/releases/tag/v1.9.6

It helps us continually improve our product. If you encounter any issues while using the new version, please do not hesitate to let us know.

Thank you once again!

For this issue, special thanks to @twankamp

PR #1399, opened 2023-08-12
feat(connection): ALPN protocols support

twankamp:
### PR Checklist

No support for setting ALPN Protocols

#### Issue Number

Example: \#916

Support for setting ALPN Protocols

- [ ] Yes
- [x] No

#### Specific Instructions


ysfscream:
@twankamp Thank you so much for the PR! We truly appreciate your contribution. However, could you provide more detailed documentation or descriptions regarding this feature update? It would be great if you could update the documents or provide usage steps and screenshots within the PR.

Additionally, is ALPN a widely adopted capability? I'm uncertain if its application is exclusive to AWS or if it's also common elsewhere.

Looking forward to your updates!

twankamp:
@ysfscream certainly

1. This feature is applicable if you are using a TLS connection (so e.g. MQTTS or WSS) and the server needs the client to specify ALPN protocol for connecting. In that case when you select SSL/TLS in the MQTTX desktop app, it will unfold extra options like SSL Secure and Certificate, and in there now also ALPN is added. Then you simply specify the desired ALPN protocol in that field alongside all other parameters you need, then click connect and it will work. Note that officially ALPN can be an array but that is an extremely niche use-case so I didn't want to mention about it to avoid confusion. However just in case if the user specifies multiple protocols in a way like p1,p2 or something like [p1, p2 ] then I made sure it will work as well.

image

2. This is a universally recognized capability, it doesn't have anything specific to do with AWS - I think the person in the feature request #916 (whom I don't know) just used it as an example. This feature is often used to improve performance if you host multiple protocols on the same port (for example a broker that supports both mqtts and wss on 443, or a reverse proxy that covers both an https and mqtts server). Even the library you use (mqtt.js) already supports it out of the box, so the only thing I did was expose it in the UI so users of the desktop app can actually use it. Other clients use it as well (e.g. https://mosquitto.org/man/mosquitto_pub-1.html), but the benefit of your app is that it has a GUI.

src/utils/mqttUtils.ts
```diff
@@ -86,6 +87,9 @@ const getClientOptions = (record: ConnectionModel): IClientOptions => {
   // SSL
   if (ssl) {
     options.rejectUnauthorized = rejectUnauthorized === undefined ? true : rejectUnauthorized
+    if (ALPNProtocols) {
+      options.ALPNProtocols = ALPNProtocols.replace(/[\[\] ]/g, '').split(',')
+    }
```
ysfscream, line 92: Sorry for later.

It seems that you're trying to handle ALPN protocols by cleaning up the string and then converting it to an array of protocols. However, I noticed a potential issue: I couldn't find any reference to ALPN handling within MQTT.js' official documentation. Can you clarify how adding ALPNProtocols to the options would affect the functioning of the MQTT client?

Have you tested this integration? Is it currently in a working state? Your insights would be valuable to understand its efficacy and compatibility with MQTT.js.

ysfscream: I apologize for the oversight. After thoroughly reviewing the `mqtt.js` source code, I did find the mentioned field.

https://github.com/mqttjs/MQTT.js/blob/3aaacff51f03d78bd4f334d30c9ceda1d86525be/src/lib/client.ts#L111

It seems there aren't any significant issues. However, I'm not very familiar with this feature. Could you provide more use cases or examples? This would help me understand better and determine how to introduce and document it in the future. Thank you again!

twankamp: @ysfscream

Yes I have tested this integration thoroughly and also had other people test my build and I can confirm it's in a fully working state.

Indeed you can see the field is an array, but as mentioned in almost all cases for an MQTT broker I expect the user to only enter one protocol. However the string cleaning takes care that most common ways for users to enter an array of protocols will also work if someone ever needs it.

I described use-cases and examples in my comment above.
The typical case is that you have a broker which supports multiple tls protocols such as mqtts and mqtt over wss on the same port, then the clients needs to select the protocol in ALPN, so the server can handle the handshake accordingly. The reason to use the same port is often that some network environments do not have port 8883 open, so instead you want to offer both wss and mqtts on port 443. The user would then enter enter 'mqtt' or 'wss' in ALPN. Other use-cases could be that you have a server setup with a reverse proxy that proxies to other services than just mqtt broker on the same port, so again you need to indicate the target service/protocol.
Another use-case is for example that you want the broker to support multiple authentication methods also on the same port like 443, e.g. client certificate for devices vs username/password for applications. So then for example the server specifies that clients have to put 'mqtt-ca' in ALPN to indicate CA authentication.

I suggest to simply document it similar as other mqtt clients do, for example:
ALPN: Configure the application layer protocol negotiation option for the TLS session. Useful for brokers that have multiple protocols available on a single port, e.g. MQTT and WebSockets.

In general it's one of those things if a user doesn't know what it is then they just leave it empty. But if the server they want to connect it needs it, that will be clearly indicated by the instructions to connect to that server, so then the user will easily be able to find where to enter it.

ysfscream (approved):

ysfscream:
@twankamp A huge thanks for your contribution! 🚀 We're excited to roll this out in our upcoming release.

src/utils/mqttUtils.ts
twankamp: Yes that is correct.

Thanks!

PR #1435, opened 2023-09-24
feat(deps): use specified node version for build

Rotzbua:

Node version for build environment is defined in the GH Actions scripts.

#### Issue Number

none

Use central config file to manage node version.

- [ ] Yes
- [x] No

#### Specific Instructions

none

#### Other information

none


Red-Asuka (approved):

Red-Asuka (changes requested):
@Rotzbua It seems that the new feature only supports `node-version-file` v2.5.0 or above. Please update the version.

https://github.com/actions/setup-node/releases/tag/v2.5.0

Red-Asuka (approved):

Swilder-M (approved):

ysfscream:
@Rotzbua
Thank you for your PR! You mentioned moving the Node version definition from the GH Actions scripts to using a centralized config file, which is a great improvement.

Before merging, I'd like to confirm one thing: Have you tested to ensure that we can still build a functional package after this change? We want to ensure this modification doesn't adversely affect our current build process.

I'm looking forward to hearing back from you!

Issue #1398, opened 2023-08-07
[Bug] An error occurs when a received message contains a topic alias


tigercl:
#### What did I do

I use MQTTX CLI to subscribe to a topic. When the server forwards a message containing the topic alias to MQTTX CLI, the CLI will print the following error log:

![1161690970213_ pic](https://github.com/emqx/MQTTX/assets/15085956/533d6215-0176-4562-b9b2-0b0912568773)

And, when I set a topic alias, and send multiple messages with the `--multiline` option, each message will contain both the topic and the topic alias:

```
mqttx pub --topic mytopic --topic-alias 1 --stdin --multiline
```

In fact, except for the first message, all subsequent messages only need to contain the topic alias.

#### Expected

MQTTX CLI should receive messages with the topic alias normally.

#### Environment

- OS: OS version, e.g. macOS Big Sur 11.1
- MQTTX version, e.g. v1.9.4

ysfscream:
Thank you for the feedback. The issue should be fixed in v1.9.7. Please feel free to test and provide additional feedback.

PR #1444, opened 2023-09-28
fix(cli):fix topic-alias-maximum error in cli connect

ni00:
### PR Checklist

1. When using --topic-alias-maximum during sub and connect, it will cause an error when receiving messages.
It's normal to use --topic-alias-maximum when pub.
2. set a topic alias, and send multiple messages with the --multiline option, each message will contain both the topic and the topic alias.

#### Issue Number

#1398 

1. fix topic-alias-maximum bug
2. only send once topic when multiline send and topic-alias is exist

- [ ] Yes
- [x] No

#### Specific Instructions

### DIFF
--- a/cli/src/lib/pub.ts
+++ b/cli/src/lib/pub.ts
@@ -95,10 +95,13 @@ const multisend = (
   const sender = new Writable({
     objectMode: true,
   })
+  let count = 0
   sender._write = (line, _enc, cb) => {
     const { topic, opts, protobufPath, protobufMessageName, format } = pubOpts
+    count++
+    let omitTopic = opts.properties?.topicAlias && count >= 2
     const publishMessage = processPublishMessage(line.trim(), protobufPath, protobufMessageName, format)
-    client.publish(topic, publishMessage, opts, cb)
+    client.publish(omitTopic ? '' : topic, publishMessage, opts, cb)
   }
 
   client.on('connect', () => {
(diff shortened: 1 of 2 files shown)


cli/src/lib/pub.ts
```diff
@@ -95,10 +95,13 @@ const multisend = (
...
     objectMode: true,
   })
+  let count = 0
   sender._write = (line, _enc, cb) => {
     const { topic, opts, protobufPath, protobufMessageName, format } = pubOpts
+    count++
+    let omitTopic = opts.properties?.topicAlias && count >= 2
     const publishMessage = processPublishMessage(line.trim(), protobufPath, protobufMessageName, format)
-    client.publish(topic, publishMessage, opts, cb)
+    client.publish(omitTopic ? '' : topic, publishMessage, opts, cb)
```
ysfscream, line 104: What is this for? If a topic alias is set, then after a line break, do we not need to enter the original input topic anymore?

ni00: We only need to send the topic and topic alias together for the first time. The second time, we can directly use the topic alias to send, and the topic is set to "". The current multi-line send will carry the topic and topic alias each time, which does not comply with the mqtt protocol specification.

ysfscream (approved):

Red-Asuka (approved):

Issue #1505, opened 2023-11-24
[Bug] Issue with Data Precision Loss in JSON Messages in MQTTX


nbptzjp:
#### What did I do

1. Used Java code to call EMQX's interface to send JSON-type messages.
2. Created two connections in MQTTX: Connection A to send MQTT messages, and Connection B to subscribe and receive messages sent by Connection A.

#### What happened

原始报文
`{"code":10271,"value":{"phoneBTMac":"11:1A:2B:3C:4D:9E","pinCode":522676,"serializable":1724442618569203714,"userId":1688888483063664642}}`
mqttx接收到的报文
`{
  "code": 10271,
  "value": {
    "phoneBTMac": "11:1A:2B:3C:4D:9E",
    "pinCode": 245405,
    "serializable": 1724442001931018200,
    "userId": 1688888483063664600
  }
}`

#### Expected

报文中的long类型数据精度丢失

#### Environment

- OS: OS version, e.g. win11
- MQTTX version, e.g.  v1.9.6

#### More detail

More detail include `log` information or screenshot

nbptzjp:
![image](https://github.com/emqx/MQTTX/assets/36099349/2054332f-da3d-4bc5-a995-0e82b6f274b2)
A链接发送报文 
{"code":10271,"value":{"phoneBTMac":"11:1A:2B:3C:4D:9E","pinCode":522676,"serializable":1724442618569203714,"userId":1688888483063664642}}
B链接分别用json和text接收报文
![image](https://github.com/emqx/MQTTX/assets/36099349/98d8d8d7-4372-4857-a91a-385e66a15dc9)




日志
[2023-11-24 14:27:24] [INFO] A链接 message arrived: message added "message_3f7676ef-896a-42d8-a2fe-4a22c95a3bc4" and added to topic: "sys/2679047356254564354/data_upload", payload: "{\n  \"code\": 10271,\n  \"value\": {\n    \"phoneBTMac\": \"11:1A:2B:3C:4D:9E\",\n    \"pinCode\": 522676,\n    \"serializable\": 1724442618569203700,\n    \"userId\": 1688888483063664600\n  }\n}" MQTT.js onMessageArrived trigger with Properties: undefined
[...]

Issue #1454, opened 2023-10-18
[Bug] Issue with returning arrays exceeding 16 digits in json format


yinft:
#### What did I do

I sent a 19 digit uuid

#### What happened

When I choose to use json format encoding to receive the payload, the client can only recognize 16 digits, and the following three digits all become 0. It’s fine if I use plaintext format. 

This problem also occurs in web applications. The browser can only recognize 16-digit number types.

#### Expected

Numeric types exceeding 16 digits in json format can be returned unchanged

#### Environment

- OS: OS version, e.g. macOS Big Sur 11.1
- MQTTX version, e.g. v1.5.0

#### More detail

More detail include `log` information or screenshot

ysfscream:
Hi, @yinft 

This issue is planned to be addressed in version 1.9.7. Currently, our JSON format does not support BigInt Number.

Based on your description, this issue seems to align with what we have documented in https://github.com/emqx/MQTTX/issues/1410 and https://github.com/emqx/MQTTX/issues/740. Thank you very much for your feedback. We'll prioritize and work on the fix immediately.

yinft:
Is it because the front end uses bigint? Now when I send the 19-digit uuid again, there will still be a problem with the last digit. I found the corresponding problem online
https://segmentfault.com/q/1010000042975309

ysfscream:
Hi, @ni00 Please help take another look at this issue, thank you.

ni00:
@ysfscream Okay, I will fix this problem this week.

ni00:
![image](https://github.com/emqx/MQTTX/assets/56543214/aa82ec91-5ac3-4950-9dee-5d28e5729b28)

Hello @yinft , could you provide me with a copy of your payload? Because in the versions of MQTTX after v1.9.7 (the current version is v1.9.9), I am unable to reproduce the bug issue you described.

yinft:
今天我也没有复现这个问题了，真的很奇怪，有可能是我代码的哪有问题吧，先关闭这个issues吧

Issue #1410, opened 2023-09-05
[Feature] Support for bigint on the JSON parsing


BinarSkugga:
#### Motivation

Support for 64bit numbers. Because JS uses a 52bit mantissa for all numbers, ids are truncated.

#### Detailed design

I know there is a couple of library in JS that do this:
https://www.npmjs.com/package/json-bigint (https://www.npmjs.com/package/@types/json-bigint)
https://www.npmjs.com/package/json-bigint-patch

#### Alternatives

Currently using the plaintext view.

#### More detail (optional)

With parsing:
![image](https://github.com/emqx/MQTTX/assets/7575628/24e8deb7-514b-45e9-bd6d-41fc4facef72)

Without:
![image](https://github.com/emqx/MQTTX/assets/7575628/5616462b-d998-4819-ad4d-9b9ca30869c5)

ysfscream:
@BinarSkugga Thank you for the detailed information and suggestions! I noticed that this issue seems related to https://github.com/emqx/MQTTX/issues/740

We will address this in upcoming versions. Thanks again for your contribution and patience!

ysfscream:
This issue has been fixed in the [MQTTX v1.9.7 release](https://github.com/emqx/MQTTX/releases/tag/v1.9.7). Thank you for your feedback.

Issue #740, opened 2021-10-15
[BUG] Large number reception in the json format causes an exception


wjsffzb:
version:1.6.2
os:windows10

oceanlvr:
More detail?

oceanlvr:
If you can modify the problem in Englis.

wjsffzb:
if MQTT response contain big number,convert to json will overflow and no hint.

PR #1507, opened 2023-11-30
fix:resolve data precision loss in JSON messages

ni00:
### PR Checklist

If the MQTT response contains a large number (e.g. 8 bytes like 47348660069859329), using JSON format will cause the accept to overflow without any indication.The functions of subscribing messages, sending messages, and input box formatting in the web end, desktop end, and CLI end all have the issue of JSON precision loss.

image1

#### Issue Number

#1505 #1410 #740  #1454 

display normally
image2

- [ ] Yes
- [x] No

#### Specific Instructions


Red-Asuka:
The path of the .nvmrc file configured in action is incorrect, can you fix it by the way?

> Error: The specified node version file at: /home/runner/work/MQTTX/MQTTX/.nvmrc does not exist

Red-Asuka:
By the way, how does json-bigint perform compared to the native method? Do you have any relevant tests to share?

ni00:
> By the way, how does json-bigint perform compared to the native method? Do you have any relevant tests to share?

I ran some code tests, and found that in the case of parsing and stringifying 10,000 times, the JSON-BigInt parser took roughly three times longer than the native JSON parser. I think this time overhead is completely acceptable.

Test Code:
```typescript
const JSONbigNative = require("json-bigint")({ useNativeBigInt: true });

// Test String
const str = `
{"code":10000,"value":{"phoneBTMac":"11:1A:2B:3C:4D:AA","pinCode":134123,"serializable":172444261856920371423423424,"userId":16888884830636646423242342}}
`;

// Native
console.time("Native");
for (let i = 0; i < 10000; i++) {
  JSON.stringify(JSON.parse(str));
}
console.timeEnd("Native");

// Bigint
console.time("Bigint");
for (let i = 0; i < 10000; i++) {
  JSONbigNative.stringify(JSONbigNative.parse(str));
}
console.timeEnd("Bigint");
```

Output:
Native: 22.733ms
Bigint: 76.5ms

Red-Asuka (approved):

ni00:
In my recent test, using json-bigint and native bigint will cause the JSON string to fail to parse floating-point numbers (because it will treat all numbers as bigint), which will lead to serious errors, so I will modify the code. Please do not merge this PR for the time being.

eg:
`SyntaxError: Cannot convert 7.2444261856920371423423424 to a BigInt`

ni00:
I fixed the previous issue and optimized the code. JavaScript has a built-in bigint type for handling long integers, but it lacks a built-in type for long floating-point numbers, so I introduced a fallback with BigNumber.

I used the native JSON library to handle MQTTX configuration files (such as user parameters, meta parameters, etc.), protobuf schema content, and script content. This is because MQTTX configuration files do not contain long integers or long floating-point numbers. Additionally, the protobuf library's validation only supports the native JavaScript number type and does not support extended bigint and bignumber types. Furthermore, the scripts use JavaScript operations and the sandbox does not have a built-in bignumber, which may result in loss of precision. Therefore, the native JSON library is sufficient for these purposes.
When the JSON file does not contain floating-point values internally, I used JSONBigInt, which is almost identical to native JSON, except that the number type becomes bigint, while the object structure remains unchanged. Conversely, I used JSONBigNumber when the JSON file contains floating-point numbers. After parsing the JSON, long integers and floating-point numbers will be represented uniformly as BigNumber objects. When stringified, floating-point numbers will remain unchanged, while long integers will be converted to scientific notation floating-point numbers (which is supported by JSON). Both JSONBigInt and JSONBigNumber only affect the payload and do not alter the specific content.

image

ysfscream:
@ni00 Sure. Thank you very much for your effort and the solution to the JSON precision issue in MQTTX. Using JavaScript's bigint and BigNumber to handle long integers and floating-point numbers is a great approach. Thank you again for your contribution! I will merge the PR and test it locally.

cli/src/utils/jsonUtils.ts
```diff
@@ -0,0 +1,24 @@
+const JSONBigNumber = require('json-bigint')
+const JSONBigInt = require('json-bigint')({ useNativeBigInt: true })
```
Red-Asuka, line 2: Can it be changed to ES6 import?

Red-Asuka (approved):

Issue #735, opened 2021-10-11
Add save button for new connection


JannemanDev:
#### Motivation

When you want to make a new connection and fill in some (but not necessarly all) fields you can't temporailty save it and  complement it later on. You can only choose Connect which is not always what you want.

Suggestion: add a save button next to connect.

GuryYu:
And edit connection needs save button too.

ysfscream:
Thanks for your feedback. It is now available at: https://github.com/emqx/MQTTX/releases/tag/v1.9.10. Welcome to try it.

PR #1610, opened 2024-03-23
feat(desktop): add save button for new & edit connection

XL-YiBai:
### PR Checklist

When you want to make a new connection and fill in some (but not necessarly all) fields you can't temporailty save it and complement it later on. You can only choose Connect which is not always what you want. The same goes for editing connections.

#### Issue Number

#735

When creating or editing connection information, the current information can be saved. Instead of clicking the `Connect` button to establish a connection.

- [ ] Yes
- [x] No

#### Specific Instructions

#### Other information

屏幕录制2024-03-25 10 19 26

### DIFF
--- a/src/lang/common.ts
+++ b/src/lang/common.ts
@@ -111,6 +111,13 @@ export default {
     ja: '削除に失敗しました',
     hu: 'Sikertelen törlés',
   },
+  saveSuccess: {
+    zh: '保存成功',
+    en: 'Save Success',
+    tr: 'Kaydetme Başarılı',
+    ja: '保存に成功しました',
+    hu: 'Sikeres mentés'
+  },
   warning: {
     zh: '提示',
     en: 'Warning',
--- a/src/views/connections/ConnectionForm.vue
+++ b/src/views/connections/ConnectionForm.vue
@@ -20,9 +20,19 @@
         <h2>{{ oper === 'create' ? $t('common.new') : $t('common.edit') }}</h2>
       </div>
       <div class="tail">
-        <a href="javascript:;" @click="save">
+        <a href="javascript:;" @click="connect" class="connect-btn">
           {{ $t('connections.connectBtn') }}
         </a>
+        <el-dropdown trigger="click" @command="handleActionCommand">
+          <a href="javascript:;">
+            <i class="el-icon-arrow-down"></i>
+          </a>
+          <el-dropdown-menu class="connection-oper-item" slot="dropdown">
+            <el-dropdown-item command="handleSave">
+              <i class="el-icon-folder"></i>{{ $t('common.save') }}
+            </el-dropdown-item>
+          </el-dropdown-menu>
+        </el-dropdown>
       </div>
     </div>
 
@@ -632,7 +642,7 @@ export default class ConnectionForm extends Vue {
   private handleCreateNewConnection(val: string) {
     if (val === 'create') {
       // reinit the form when page jump to creation page
-      this.record = _.cloneDeep(this.defaultRecord)
+      this.initRecord()
     }
   }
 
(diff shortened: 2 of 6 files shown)


Red-Asuka:
@XL-YiBai Can you provide a screenshot after the modification?

Red-Asuka (approved):

Red-Asuka:
Looking good, can we also add this feature to the MQTTX Web?

XL-YiBai:
> @XL-YiBai Can you provide a screenshot after the modification?

I just uploaded a demonstration video. 😆
Create a new connection also has the same effect

XL-YiBai:
> Looking good, can we also add this feature to the MQTTX Web?

Of course, I will complete it in my free time.

ysfscream:
@XL-YiBai 

Thank you for your contribution; it indeed looks like a valuable enhancement. However, for MQTTX, modifying parameters and saving without immediate connection isn't a frequent operation. Thus, we propose integrating this functionality via a dropdown menu to maintain a cleaner interface. Specifically, we can have the "Connect" button for standard operations on the left, and a dropdown menu on the right with an option for "Save Only". This approach benefits us by:

1. **Keeping the UI Clean**: Reducing button clutter for a more focused user experience.
2. **Intuitive Functionality**: The dropdown menu clearly indicates additional actions without overwhelming the main interface.
3. **Streamlining Common Processes**: Ensures a smooth flow for the most common action—connecting—while offering flexibility for different user needs.

### Resources:

https://element-plus.org/zh-CN/component/dropdown.html

image

the Right icon: 

image

the UI design:

image

XL-YiBai:
@ysfscream Thank you for your valuable suggestion and detailed design proposal. I will incorporate the suggested changes into the code.

XL-YiBai:
image

src/views/connections/ConnectionForm.vue
```diff
@@ -671,58 +681,108 @@ export default class ConnectionForm extends Vue {
...
+        if (!valid) {
+          resolve(false)
           return
         }
-      }
-      if (this.oper === 'create') {
-        // create a new connection
-        res = await connectionService.create({
+
+        const data = { ...this.record }
```
ysfscream, line 692: If this is merely for data validation without involving the original data processing, there's no need to perform object destructuring, which would add unnecessary performance overhead.

src/views/connections/ConnectionForm.vue
```diff
@@ -671,58 +681,108 @@ export default class ConnectionForm extends Vue {
...
+    this.$message.success(this.$tc('common.saveSuccess'))
+    this.$emit('refresh')
+    this.handleBack(id)
+  }
+
+  private async saveData() {
+    const data = { ...this.record }
+    const { connectionService } = useServices()
+    if (this.oper === 'create') {
+      localStorage.setItem('newConnectionRecord', JSON.stringify(data))
```
ysfscream, line 728: I didn't quite understand this operation. Why store it in localStorage? Shouldn't it be about creating a new connection to the database?

src/views/connections/ConnectionForm.vue
```diff
@@ -671,58 +681,108 @@ export default class ConnectionForm extends Vue {
...
+    const data = (await this.saveData())!
+
+    if (this.oper === 'create') {
+      res = await connectionService.create({
+        ...data,
+        properties: emptyToNull(data.properties),
+        createAt: time.getNowDate(),
+        updateAt: time.getNowDate(),
+      })
+      localStorage.removeItem('newConnectionRecord')
```
ysfscream, line 764: This seems like an unnecessary step. Are you planning to delete the locally saved new connection and create it during the connect process?

src/views/connections/ConnectionForm.vue
XL-YiBai: My understanding of the requirement raised by this [issue](https://github.com/emqx/MQTTX/issues/735) is that the configuration can be temporarily saved when creating a new connection, and the saved configuration can be displayed when clicking on the new connection again. 
Creating a new connection to the database can also achieve temporary storage, but in this way, the default configuration will always be displayed every time you click on the new connection. 
So I stored it in localStorage and then removed it when the user clicked `Connect` button.

XL-YiBai: Sorry. It seems that I misunderstood, I will modify the code.

ysfscream: Okay, I understand what you mean now. Actually, it doesn't need to be so complicated. Let's simplify the requirements. We just need to create or update a connection's information. We can fully reuse the previous code for saving the connection and navigating to the connection. In the "save" action, all we need to do is not to navigate to the connection page and perform the automatic connection operation. Do you understand what I mean? 

```js
// update ActiveConnection & connect
if (res && res.id) {
  this.changeActiveConnection({
    id: res.id,
    client: {
      connected: false,
    },
  });
  this.$emit('connect');
  this.$router.push(`/recent_connections/${res.id}`);
} else {
  this.$message.error(msgError);
  this.$log.error(msgError);
}
```

We just don't need to emit a 'connect': `  this.$emit('connect');`; in my understanding, everything else is the same.

web/src/utils/api/connection.ts
```diff
@@ -55,4 +55,16 @@ export const updateConnectionPushProp = (id: string, properties: MessageModel['p
...
+  return db.get('newConnectionRecord')
+}
+
+export const createNewConnectionRecord = (data: ConnectionModel) => {
+  db.set('newConnectionRecord', data)
+}
+
+export const removeNewConnectionRecord = () => {
+  db.unset('newConnectionRecord')
+}
```
ysfscream, line 68: From the desktop requirements just mentioned, it's the same for the Web. Just create a new connection without automatically connecting it.

Therefore, these unnecessary storage operations can be removed.

src/views/connections/ConnectionForm.vue
XL-YiBai: Yes, I understand. I have made the problem more complicated. 👀

src/views/connections/ConnectionForm.vue
```diff
@@ -671,58 +681,96 @@ export default class ConnectionForm extends Vue {
...
+    const valid = await this.validateForm()
+    if (!valid) {
+      return
+    }
+    const { id } = this.$route.params
+    await this.saveData()
+    this.$message.success(this.$tc('common.saveSuccess'))
+    this.$emit('refresh')
+    this.handleBack(id)
+  }
```
ysfscream, line 722: Your code appears to be unclean for the following reasons:

1. The `validateForm` function is called in both the save and connect functions, resulting in duplication within the `handleSave` and `connect` functions.
2. From my perspective, the only difference seems to be a `$emit('refresh')` in one case and a `$emit('connect')` in the other. Is that correct?

ysfscream: We need a function, `handleSave`, to handle all the save logic, determining through a parameter `type: connect | save` whether it's a connection or just a save. Then, in different action events, call this `handleSave`, handling different logic according to the type. This would make the code simpler, right?

ysfscream: Same in the Web version code.

ni00 (changes requested):

src/lang/common.ts
```diff
@@ -111,6 +111,13 @@ export default {
     ja: '削除に失敗しました',
     hu: 'Sikertelen törlés',
   },
+  saveSuccess: {
+    zh: '保存成功',
+    en: 'Save Success',
```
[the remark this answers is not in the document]
ni00: It seems that it can be changed or not.

src/views/connections/ConnectionForm.vue
XL-YiBai: Yes, the code here  seems to have some redundancies. After making the suggested modifications, the code indeed appears more concise. Thank you very much for the insightful suggestion! 🍻

src/lang/common.ts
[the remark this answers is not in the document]
XL-YiBai: Thank you for your suggestion.🍻  But I think the translation here should be consistent with other places, such as `Edit Success` and `Delete Success`.

src/views/connections/ConnectionForm.vue
```diff
@@ -671,43 +681,70 @@ export default class ConnectionForm extends Vue {
...
+      return
+    }
+
+    const res = await this.saveData()
+    if (type === 'save') {
+      const { id } = this.$route.params
+      this.$emit('refresh')
+      this.handleBack(id)
+      this.$message.success(this.$tc('common.saveSuccess'))
+    } else {
```
ysfscream, line 745: Even with save, there's still a chance of failure after saving the data...

Do not add error messages for creation or editing failures; only include them in the connection section.

src/views/connections/ConnectionForm.vue
```diff
@@ -722,7 +759,13 @@ export default class ConnectionForm extends Vue {
...
         this.$log.error(msgError)
       }
-    })
+    }
+  }
+
+  private handleActionCommand(command: string) {
+    if (command === 'handleSave') {
+      this.handleSave('save')
+    }
```
ysfscream, line 768: It's recommended to use `save` for commands, as `handleSave` sounds more like a function name.

Consider future scalability.

```suggestion
  private handleActionCommand(command: 'save') {
    if (command === 'save') {
      this.handleSave('save')
    }
```

ysfscream (approved):
LGTM! Thanks

Issue #1370, opened 2023-07-20
MQTTX shows blank screen after upgrading to 1.9.4


AdonLuo:
#### What did I do

Unable to load properly, shows blank screen.

Menu can be force loaded to show, but turns blank again after 1 second.

#### Environment

- OS: Win10
- MQTTX version：1.9.4
![1689816886779](https://github.com/emqx/MQTTX/assets/19642911/1b13303b-b807-4d45-a1e5-3ec58f8d453d)

ysfscream:
Welcome to try the new MQTTX release at https://github.com/emqx/MQTTX/releases/tag/v1.9.5 If problems persist, please let us know and we will promptly address them.

AdonLuo:
1.9.5 仍然出现这个问题，之前不会的，今天使用突然又这样了，强制加载显示正常，一会会就全白了 @ysfscream 
![image](https://github.com/emqx/MQTTX/assets/19642911/2ce718a2-f0b9-4f08-baac-5b2dea7b0272)

ysfscream:
- Use two ways to handle the white screen situation at https://github.com/emqx/MQTTX/releases/tag/v1.10.0
    - By resetting the database
    - Optimizing the display of large messages in the message list

Thanks for your feedback. If there is any problem, I will reopen this issue.

PR #1627, opened 2024-04-11
feat(desktop): add database reset page to fix blank screen from corrupted file

DM1-1:
### PR Checklist

The current behavior is that the application displays a blank screen when the database file is corrupted, preventing users from interacting with the application.

#### Issue Number

#1370

The new behavior is that upon detecting a database corruption, the application will redirect the user to an error page where they can initiate a database reset. This resolves the white screen issue and allows the user to restore the application to a working state.

- [ ] Yes
- [x] No

#### Specific Instructions

If you wish to conduct testing, you can do so by altering the fields within the `MQTTX.db` file in the application data to make them non-conforming to the schema.

#### Other information

The UI and UX for the error page and database reset feature are still a work in progress and will be refined in future updates.
This change has been tested on macOS and Windows platforms to confirm that it resolves the white screen issue caused by database corruption.

### DIFF
--- a/src/background.ts
+++ b/src/background.ts
@@ -4,6 +4,7 @@ import { app, protocol, BrowserWindow, ipcMain, shell, Menu } from 'electron'
 import { createProtocol } from 'vue-cli-plugin-electron-builder/lib'
 import installExtension, { VUEJS_DEVTOOLS } from 'electron-devtools-installer'
 import { quitAndRenameLogger } from './utils/logger'
+import rebuildDatabase from './utils/rebuildDatabase'
 import { createUpdateWindow, autoDownload } from './main/updateDownloader'
 import { getCurrentLang } from './main/updateChecker'
 import getMenuTemplate from './main/getMenuTemplate'
@@ -14,6 +15,7 @@ import { onSystemThemeChanged } from './main/systemTheme'
 import useConnection, { initOptionModel } from '@/database/useConnection'
 import useServices from '@/database/useServices'
 import { dialog } from 'electron'
+import ORMConfig from './database/database.config'
 import version from '@/version'
 
 declare const __static: string
@@ -92,6 +94,20 @@ function handleIpcMessages() {
   ipcMain.on('insertCodeToEditor', (event, ...args) => {
     event.sender.send('insertCodeToEditor', ...args)
   })
+  ipcMain.on('rebuildDatabase', (event) => {
+    try {
+      rebuildDatabase(ORMConfig.database as string)
+      app.relaunch()
+      app.exit()
+    } catch (error) {
+      dialog.showMessageBox({
+        type: 'error',
+        title: 'Rebuild Database Error',
+        message: 'An error occurred while rebuilding the database.',
+        detail: error.message
+      })
+    }
+  })
 }
 
 // handle event when APP quit
@@ -106,30 +122,41 @@ function beforeAppQuit() {
 
 async function createWindow() {
   // Init tables and connect to local database.
-  await ConnectionInit({ doMigrations: true } as initOptionModel)
-  const { settingService } = useServices()
-  await settingService.set()
-  const setting = await settingService.get()
-  if (setting) {
-    theme = setting.currentTheme
-    autoCheckUpdate = setting.autoCheck
-    syncOsTheme = setting.syncOsTheme
-    windowSize.height = setting.height
-    windowSize.width = setting.width
+  try {
+    await ConnectionInit({ doMigrations: true } as initOptionModel)
+    const { settingService } = useServices()
+    await settingService.set()
+    const setting = await settingService.get()
+    if (setting) {
+      theme = setting.currentTheme
+      autoCheckUpdate = setting.autoCheck
+      syncOsTheme = setting.syncOsTheme
+      windowSize.height = setting.height
+      windowSize.width = setting.width
+      //@ts-ignore
+      global.sharedData = {
+        currentTheme: setting.currentTheme,
+        currentLang: setting.currentLang,
+        autoCheck: setting.autoCheck,
+        autoResub: setting.autoResub,
+        maxReconnectTimes: setting.maxReconnectTimes,
+        syncOsTheme: setting.syncOsTheme,
+        multiTopics: setting.multiTopics,
+        jsonHighlight: setting.jsonHighlight,
+        enableCopilot: setting.enableCopilot,
+        openAIAPIKey: setting.openAIAPIKey,
+        model: setting.model,
+        logLevel: setting.logLevel,
+      }
+    }
+  } catch (error) {
+    console.error('ConnectionInit error:', error)
     //@ts-ignore
     global.sharedData = {
-      currentTheme: setting.currentTheme,
-      currentLang: setting.currentLang,
-      autoCheck: setting.autoCheck,
-      autoResub: setting.autoResub,
-      maxReconnectTimes: setting.maxReconnectTimes,
-      syncOsTheme: setting.syncOsTheme,
-      multiTopics: setting.multiTopics,
-      jsonHighlight: setting.jsonHighlight,
-      enableCopilot: setting.enableCopilot,
-      openAIAPIKey: setting.openAIAPIKey,
-      model: setting.model,
-      logLevel: setting.logLevel,
+      connectDatabaseFailMessage: error.message,
+      currentTheme: 'light',
+      currentLang: 'en',
+      syncOsTheme: false,
     }
   }
   // Create the browser window.
@@ -156,7 +183,8 @@ async function createWindow() {
     }
   })
   // Menu Manger
-  const templateMenu = getMenuTemplate(win, setting?.currentLang)
+  // @ts-ignore
+  const templateMenu = getMenuTemplate(win, global.sharedData.currentLang)
   menu = Menu.buildFromTemplate(templateMenu)
   Menu.setApplicationMenu(menu)
 
--- a/src/components/DatabaseError.vue
+++ b/src/components/DatabaseError.vue
@@ -1,0 +1,87 @@
+<template>
+  <div class="connect-database-error-page">
+    <div class="page-content">
+      <p class="info">A corruption has been detected in the database file, preventing the software from launching properly. To fix this, please click 'Rebuild Database'.</p>
+      <p class="error">Error - {{ connectDatabaseFailMessage }}</p>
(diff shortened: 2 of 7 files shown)


src/components/DatabaseError.vue
```diff
@@ -0,0 +1,45 @@
...
+    border: 1px solid #37dc85;
+  }
+}
+
+.error-content {
+  font-size: 24px;
+  font-weight: 500;
+  text-align: center;
+}
+</style>
```
ysfscream, line 45: Add a new blank line at the end of the file.

src/utils/rebuildDatabase.ts
```diff
@@ -0,0 +1,10 @@
+const fs = require('fs')
+
+export const rebuildDatabase = (filePath: string) =>  {
+  try {
+    fs.writeFileSync(filePath, '')
+    console.log('Rebuild Database Success!')
+  } catch (error) {
+    throw new Error('Rebuild Database Failed: ' + error)
+  }
+}
```
ysfscream, line 10: Add a new blank line at the end of the file.

src/utils/rebuildDatabase.ts
```diff
@@ -0,0 +1,10 @@
+const fs = require('fs')
+
+export const rebuildDatabase = (filePath: string) =>  {
```
ysfscream, line 3: Recommend using the following export and import method:

```js
import rebuildDatabase from './utils/rebuildDatabase'
```

since we no longer need additional utility functions in this file, right?

src/store/getter.ts
```diff
@@ -21,6 +21,7 @@ const getters = {
   isPrismButtonAdded: (state: State) => state.app.isPrismButtonAdded,
   logLevel: (state: State) => state.app.logLevel,
   showConnectionList: (state: State) => state.app.showConnectionList,
+  connectionDataBaseFail: (state: State) => state.app.connectionDataBaseFail,
```
ysfscream, line 24:
```suggestion
  connectDatabaseFail: (state: State) => state.app.connectDatabaseFail,
```

ysfscream: DataBase -> Database

src/components/DatabaseError.vue
```diff
@@ -0,0 +1,45 @@
+<template>
+  <div class="error-page">
+    <div class="error-content">
+      <p>The database file is corrupted, please reset the database file to fix the issue.</p>
+    </div>
+    <el-button class="primary-btn" @click="rebuildDatabase">
+      reset database
```
Red-Asuka, line 7: Rebuild Database

src/components/DatabaseError.vue
```diff
@@ -0,0 +1,45 @@
+<template>
+  <div class="error-page">
+    <div class="error-content">
+      <p>The database file is corrupted, please reset the database file to fix the issue.</p>
```
Red-Asuka, line 4: We have detected a corruption in the database file, which is preventing the software from launching properly. To resolve this issue, please click on 'Rebuild Database'. Be aware that this process will result in the loss of all original data.

Red-Asuka:
@ysfscream Do you think it is necessary to add a second confirmation popup and button loading?

ysfscream:
> @ysfscream Do you think it is necessary to add a second confirmation popup and button loading?

@Red-Asuka  Of course, this is an operation that requires user consent and a secondary confirmation. it's a good suggesion.

However, I'm suddenly wondering if it's possible to recover some data from the damaged database before rebuilding the database file?

DM1-1:
> > @ysfscream Do you think it is necessary to add a second confirmation popup and button loading?
> 
> @Red-Asuka Of course, this is an operation that requires user consent and a secondary confirmation. it's a good suggesion.
> 
> However, I'm suddenly wondering if it's possible to recover some data from the damaged database before rebuilding the database file? @DM1-1

When displaying this page, the database has been unable to connect successfully. I am uncertain whether it is possible to retrieve information from the database and filter out the data that caused the connection failure.

ysfscream:
@DM1-1 YES, i think so. I mean, we can minimize the loss of user data as much as possible. So, maybe we can do some more technical research? What do you think? Even if we can't recover all the data, we could discard the messages but keep some simple information that can be appropriately located, including connection details.

For example, we could directly read the database file, or look online for other methods. If it's completely damaged and unreadable, this method won't work, and we can display this page as is.

src/components/DatabaseError.vue
ysfscream: This is a good prompt. For the warning part, we can use an Alert component from https://element.eleme.cn/#/zh-CN/component/alert.

ysfscream: And It would be even better to display the error information caught here. Mark RED.

ysfscream: It would be better if the database rebuilding process could be asynchronous so that we can display the loading status of the rebuild on this page.

src/utils/rebuildDatabase.ts
DM1-1: Yes, I should pay attention to this detail.

src/components/DatabaseError.vue
DM1-1: Has MQTTX not done code formatting?

ysfscream: /fix: I'm not sure, but from what I understand, the code formatting is triggered after the git commit.

DM1-1:
Kapture 2024-04-15 at 17 03 28

src/utils/rebuildDatabase.ts
```diff
@@ -0,0 +1,12 @@
+import fs from 'fs';
```
ysfscream, line 1:
```suggestion
import fs from 'fs'
```

src/utils/rebuildDatabase.ts
```diff
@@ -0,0 +1,12 @@
...
+
+const rebuildDatabase = (filePath: string) =>  {
+  try {
+    fs.writeFileSync(filePath, '')
```
ysfscream, line 5: **Recommendation:** Update `rebuildDatabase` to use asynchronous `fs.writeFile` with Promise for better error handling and to prevent blocking the main process.

**Example Change:**
```javascript
const rebuildDatabase = (filePath: string) => {
  return new Promise((resolve, reject) => {
    fs.writeFile(filePath, '', (error) => {
      if (error) {
        console.error('Error rebuilding database:', error);
        reject(new Error('Rebuild failed: ' + error));
      } else {
        console.log('Database rebuilt successfully.');
        resolve();
      }
    });
  });
}
```

src/background.ts
```diff
@@ -92,6 +94,20 @@ function handleIpcMessages() {
   ipcMain.on('insertCodeToEditor', (event, ...args) => {
     event.sender.send('insertCodeToEditor', ...args)
   })
+  ipcMain.on('rebuildDatabase', (event) => {
+    try {
+      rebuildDatabase(ORMConfig.database as string)
```
ysfscream, line 99: same here, try using `await rebuildDatabase(xx)`. If nothing goes wrong, we can relaunch the app.

src/components/DatabaseError.vue
```diff
@@ -0,0 +1,85 @@
+<template>
+  <div class="connect-database-error-page">
+    <div>
+      <p>A corruption has been detected in the database file, preventing the software from launching properly. To fix this, please click 'Rebuild Database'.</p>
```
ysfscream, line 4: this description needs a min-width of 60% or 70%.

ysfscream: add a margin-bottom: 14px or 16px

src/components/DatabaseError.vue
```diff
@@ -0,0 +1,85 @@
...
+      <span>
+        <strong>⚠️ Warning</strong>
+        <br>
+        Proceed with database rebuild now? All data will be lost and cannot be undone.
+      </span>
+      <span slot="footer" class="dialog-footer">
+        <el-button @click="showDialog = false">Cancel</el-button>
+        <el-button type="primary" @click="confirmRebuild">Confirm</el-button>
+      </span>
+    </el-dialog>
```
ysfscream, line 26: just using the MessageBox `$confirm` component is sufficient; there's no need to introduce a dialog and increase complexity.

src/components/DatabaseError.vue
```diff
@@ -0,0 +1,85 @@
...
+<style lang="scss">
+.connect-database-error-page {
+  display: flex;
+  flex-direction: column;
+  justify-content: center;
+  align-items: center;
+  height: 100vh;
+  gap: 40px;
+  .error-content {
+    color: red;
```
ysfscream, line 62: using the `var(--color-minor-red)` not the `red`

src/utils/rebuildDatabase.ts
```diff
@@ -0,0 +1,12 @@
...
+    console.log('Rebuild Database Success!')
```
ysfscream, line 6: This is what I'm particularly concerned about, but I forgot to include it in the PRD document. I want to confirm if we can introduce our logging tool to record every step in our log files?

src/background.ts
```diff
@@ -92,6 +94,20 @@ function handleIpcMessages() {
...
+    try {
+      await rebuildDatabase(ORMConfig.database as string)
+      app.relaunch()
+      app.exit()
+    } catch(error) {
+      dialog.showMessageBox({
+        type: 'error',
+        title: 'Rebuild Database Error',
+        message: 'An error occurred while rebuilding the database.',
+        detail: error as string
```
Red-Asuka, line 107: Use `error.message` to get the error message instead of `as string`

src/background.ts
```diff
@@ -106,30 +122,41 @@ function beforeAppQuit() {
...
-      autoResub: setting.autoResub,
-      maxReconnectTimes: setting.maxReconnectTimes,
-      syncOsTheme: setting.syncOsTheme,
-      multiTopics: setting.multiTopics,
-      jsonHighlight: setting.jsonHighlight,
-      enableCopilot: setting.enableCopilot,
-      openAIAPIKey: setting.openAIAPIKey,
-      model: setting.model,
-      logLevel: setting.logLevel,
+      connectDatabaseFailMessage: error as string,
```
Red-Asuka, line 156: Use `error.message` to get the error message instead of `as string`

DM1-1:
image

Red-Asuka (approved):

PR #1629, opened 2024-04-19
feat(cli): support for reading and writing messages from/to files

DM1-1:
### PR Checklist

The current behavior of the CLI does not support reading and writing messages from/to files.

The new behavior introduced by this PR allows the CLI to read messages from and write messages to files. This enhancement includes support for binary format for message representation.

- [ ] Yes
- [ ] No (Assuming the new feature is backward compatible and does not break existing functionality)

#### Specific Instructions

- Ensure that the file reading and writing functions are thoroughly tested with various file sizes and binary formats.
- Verify that the new feature does not introduce any performance degradation in the CLI.
- Refactoring the previous code related to saving/loading configuration into a new utility file for handling file operations.


DM1-1:
base64/hex
image
___
json/cbor
image
___
binary
image
___
loading for pub
image
___
--file-save
image
___
--file-write
image
image

DM1-1:
save & load config
image

cli/src/types/global.d.ts
```diff
@@ -81,6 +82,8 @@ declare global {
     topic: string[]
     qos?: QoS[]
     // properties of MQTT 5.0
+    fileWrite?: string
+    fileSave?: string
```
Red-Asuka, line 86: `OmitPublishOptions` and `OmitSubscribeOptions` need to exclude the newly added options.

ysfscream:
@DM1-1 GOOD JOB 👍

But I have a question: if we have successfully written it into the file, is it still necessary to print the information to the console? @Red-Asuka

cli/src/lib/pub.ts
```diff
@@ -162,17 +163,47 @@ const pub = (options: PublishOptions) => {
...
+    const filePath = processPath(options.fileRead!)
+    if (!filePath) {
+      signale.error('File path is required when reading from file.')
+      process.exit(1)
+    }
+
+    try {
+      signale.pending('Reading file...')
+      let bufferData = readFile(filePath)
+      if (bufferData.length >= 256 * 1024 * 1024) {
```
ysfscream, line 176: Can we set `256 * 1024 * 1024` as a more readable constant?

cli/src/index.ts
```diff
@@ -302,6 +306,14 @@ export class Commander {
         '--config [PATH]',
         'load the parameters from the local configuration file, which supports json and yaml format, default path is ./mqttx-cli-config.json',
       )
+      .option(
+        '--file-write <PATH>',
+        'write the received message to the file',
```
ysfscream, line 311:
```suggestion
        'append received messages to a specified file',
```

easier to understand?

Red-Asuka:
@ysfscream I think `--file-save` does not need to print the specific message content, just output the path of the saved file. `--file-write` is generally used for continuous writing of readable text, it needs to output specific message content, but you can adjust the order, first output the message content, then prompt that the file was successfully written.

cli/src/index.ts
```diff
@@ -302,6 +306,14 @@ export class Commander {
...
+      )
+      .option(
+        '--file-save <PATH>',
+        'save the received message to the file',
```
ysfscream, line 315:
```suggestion
        'save each received message to a new file',
```

cli/src/lib/pub.ts
```diff
@@ -162,17 +163,47 @@ const pub = (options: PublishOptions) => {
...
+  const handleFileRead = async () => {
+    const filePath = processPath(options.fileRead!)
+    if (!filePath) {
+      signale.error('File path is required when reading from file.')
+      process.exit(1)
+    }
+
+    try {
+      signale.pending('Reading file...')
+      let bufferData = readFile(filePath)
```
ysfscream, line 175:
```suggestion
      const bufferData = readFile(filePath)
```

Since there is no place to change the buffer data, should we use const instead of let?

cli/src/types/global.d.ts
Red-Asuka: Add to the end, or put it in front of the comment.

cli/src/utils/binaryFormats.ts
```diff
@@ -0,0 +1,13 @@
+const supportedBinaryFormatsForMQTT = [
+  '.png', '.jpg', '.jpeg', // Image file formats, common for simple media sharing
+  '.mp4', '.avi', '.mov', // Video file formats, possibly used for surveillance cameras and other devices
+  '.mp3', '.wav', // Audio file formats, suitable for voice communication or alarm systems
+  '.zip', '.gz', // Compressed file formats, suitable for firmware updates or configuration file transfers
+  '.bin', // Firmware files or other binary data
+]
```
Red-Asuka, line 7: I used GPT to complete it, but we need to check if there are any errors.

```ts
const supportedBinaryFormatsForMQTT: [
    '.png', '.jpg', '.jpeg', '.gif', '.bmp', '.ico', '.tif', '.tiff', // Image file formats
    '.mp4', '.avi', '.mov', '.mkv', '.flv', '.wmv', '.mpeg', '.3gp', // Video file formats
    '.mp3', '.wav', '.flac', '.aac', '.ogg', '.wma', '.m4a', '.m4p', // Audio file formats
    '.zip', '.gz', '.rar', '.tar', '.7z', '.bz2', '.xz', '.jar', // Compressed file formats
    '.bin', '.hex', '.exe', '.dll', '.so', '.dmg', '.iso', '.img', // Binary file formats
    '.pdf', '.epub', // Binary document file formats
]
```

cli/src/utils/binaryFormats.ts
```diff
@@ -0,0 +1,13 @@
...
+
+const isSupportedBinaryFormatForMQTT = (fileExtname: string) => {
+  return supportedBinaryFormatsForMQTT.includes(fileExtname)
```
Red-Asuka, line 10: Use `endsWith`.

cli/src/utils/config.ts
```diff
@@ -81,14 +38,15 @@ const saveConfig = (
...
 ) => {
   try {
-    const filePath = processPath(opts.save!)
-    let data: Config = {}
+    const filePath = processPath(opts.save!, defaultPath)
+    let data: Config | string = {}
     data[commandType] = removeUselessOptions(opts)
     if (fileExists(filePath)) {
-      const config = readFile(filePath)
+      const config: Config = JSON.parse(readFile(filePath).toString())
```
Red-Asuka, line 45: The configuration file read here may be in `YAML` format and should not be directly parsed using `JSON.parse`.

cli/src/lib/pub.ts
```diff
@@ -162,17 +163,47 @@ const pub = (options: PublishOptions) => {
...
+      if (bufferData.length >= 256 * 1024 * 1024) {
+        signale.error('File size exceeds 256MB, not supported')
```
ysfscream, line 177:
```suggestion
        signale.error('File size over 256MB not supported by MQTT.')
```

Isn't it better to let users know why we cannot handle data larger than 256MB?

cli/src/lib/sub.ts
```diff
@@ -4,16 +4,22 @@ import { parseConnectOptions, parseSubscribeOptions, checkTopicExists } from '..
...
   protobufPath?: string,
   protobufMessageName?: string,
   format?: FormatType,
-): string => {
+): string | Buffer => {
   let message: string | Buffer = payload
+
+  if(format === 'binary') {
+    return payload
+  }
```
ysfscream, line 22: Something seems off here? If we directly return the value in binary format, will users be unable to use the `binary` format and protobuf together? Although few users may use it this way, and we added a `binary` format for file handling, we shouldn't disrupt the original workflow. The `protobuf -> format(binary) -> buffer` data process should remain valid.

cli/src/lib/sub.ts
```diff
@@ -92,7 +98,47 @@ const sub = (options: SubscribeOptions) => {
   })
 
   client.on('message', (topic, payload, packet) => {
-    const { format, protobufPath, protobufMessageName } = options
+    const { format, protobufPath, protobufMessageName, fileSave, fileWrite } = options
+
+    let receivedMessage: string | Buffer = ''
+    if((fileSave && !fileWrite) || (!fileSave && fileWrite)) {
```
ysfscream, line 104: I think we should convert this condition into a variable for better readability and understanding.

cli/src/lib/sub.ts
```diff
@@ -92,7 +98,47 @@ const sub = (options: SubscribeOptions) => {
...
+      let savePath = ''
+      if(fileSave) {
+        savePath = processPath(fileSave)
+        savePath = findMaxNumberedFileName(savePath)
```
ysfscream, line 108:
```suggestion
       savePath = findMaxNumberedFileName(processPath(fileSave))
```

ysfscream: This is for reference only, depending on personal preference.

cli/src/lib/sub.ts
```diff
@@ -92,7 +98,47 @@ const sub = (options: SubscribeOptions) => {
...
+        signale.warn('Please use the --format binary option for handling binary files')
+      } else {
+        receivedMessage = processReceivedMessage(payload, protobufPath, protobufMessageName, format)
+      }
+
+      console.table({
+        topic,
+        size: Buffer.byteLength(receivedMessage),
+        savePath,
+      })
```
ysfscream, line 128: This isn't for debugging, right? If we need this table info, we should use `import Table from 'cli-table3'`.

cli/src/lib/sub.ts
```diff
@@ -92,7 +98,47 @@ const sub = (options: SubscribeOptions) => {
...
+        signale.error('File path is required when write to file')
+        process.exit(1)
+      }
+
+      if(!format && isSupportedBinaryFormatForMQTT(getPathExtname(savePath))) {
+        receivedMessage = processReceivedMessage(payload, protobufPath, protobufMessageName, 'binary')
+        signale.warn('Please use the --format binary option for handling binary files')
+      } else {
+        receivedMessage = processReceivedMessage(payload, protobufPath, protobufMessageName, format)
+      }
```
ysfscream, line 122:
```js
let messageFormat = format;
if (!format && isSupportedBinaryFormatForMQTT(getPathExtname(savePath))) {
    signale.warn('Please use the --format binary option for handling binary files')
    messageFormat = 'binary'
}

receivedMessage = processReceivedMessage(payload, protobufPath, protobufMessageName, messageFormat)
```

Is this a bit clearer?

cli/src/utils/fileUtils.ts
```diff
@@ -0,0 +1,86 @@
...
+}
+
+export const findMaxNumberedFileName = (filePath: string): string => {
+  if(!fileExists(filePath))
+    return filePath
+
+  const dir = path.dirname(filePath)
+  const baseNameWithoutExt = path.basename(filePath, path.extname(filePath))
+  const ext = getPathExtname(filePath)
+  const files = fs.readdirSync(dir)
```
ysfscream, line 70: Use a try-catch block here to handle any potential errors.

cli/src/utils/fileUtils.ts
```diff
@@ -0,0 +1,86 @@
...
+export const appendFile = (filePath: string, data: string | Buffer): void => {
+  try {
+    fs.appendFileSync(filePath, `${data}\n`)
+  } catch (error) {
+    signale.error(error)
+    process.exit(1)
+  }
+}
+
+export const findMaxNumberedFileName = (filePath: string): string => {
```
ysfscream, line 63: The current name may lead to misunderstandings that it is merely for finding the maximum number in existing file names.

- createNextNumberedFileName
- generateNewVersionedFileName
- getNextVersionedFilePath

Is it easier to understand the function?

cli/src/utils/fileUtils.ts
```diff
@@ -0,0 +1,86 @@
...
+  return fileExtension === '.yaml' || fileExtension === '.yml'
+}
+
+export const convertStringToData = (data: string, isYaml: boolean): Config => {
+  return isYaml ? YAML.load(data) : JSON.parse(data)
+}
+
+export const convertDataToString = (data: Config, isYaml: boolean): string => {
+  return isYaml ? YAML.dump(data) : JSON.stringify(data, null, 2)
+}
```
ysfscream, line 34: These two functions seem to be for handling `yaml` and `JSON` data, and their func names do not represent a general encoding and decoding process. Need update.

Red-Asuka (approved):

PR #1634, opened 2024-04-23
feat(cil): support for reading messages from files in bench

DM1-1:
### PR Checklist

The bench feature does not support reading messages from files.

This PR adds support for reading messages from files in the bench feature, allowing users to send messages from file sources.

- [ ] Yes
- [x] No

### DIFF
--- a/cli/src/lib/pub.ts
+++ b/cli/src/lib/pub.ts
@@ -150,6 +150,27 @@ const multisend = (
   })
 }
 
+const handleFileRead = (filePath: string) => {
+  if (!filePath) {
+    signale.error('File path is required when reading from file.')
+    process.exit(1)
+  }
+
+  try {
+    basicLog.fileReading()
+    const bufferData = readFile(filePath)
+    if (bufferData.length >= MQTT_SINGLE_MESSAGE_BYTE_LIMIT) {
+      signale.error('File size over 256MB not supported by MQTT.')
+      process.exit(1)
+    }
+    basicLog.fileReadSuccess()
+    return bufferData
+  } catch(error) {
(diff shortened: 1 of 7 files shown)


cli/src/lib/pub.ts
```diff
@@ -237,10 +237,37 @@ const multiPub = async (commandType: CommandType, options: BenchPublishOptions |
     topic,
     clientId,
     message,
+    fileRead,
     verbose,
     maximumReconnectTimes,
   } = options
 
+  const handleFileRead = () => {
```
Red-Asuka, line 245: Can't this part of the logic be shared with pub?

ysfscream: YES. This part seems a bit redundant.

ysfscream (approved):

cli/src/lib/sub.ts
```diff
@@ -57,6 +71,8 @@ const sub = (options: SubscribeOptions) => {
 
   const outputModeClean = outputMode === 'clean'
 
+  options.format = handleDefaultBinaryFile(options.format, options.fileSave! || options.fileWrite!)
```
Red-Asuka, line 74: fileSave and fileWrite are optional fields, why can we be sure they exist here?

Red-Asuka: The outer layer should be added with an if statement to check whether there is fileSave and fileWrite, or the filePath of the handleDefaultBinaryFile function should be changed to optional, and checked within the function.

DM1-1: Because the validation for checking null values has already passed.

Red-Asuka (approved):

PR #1642, opened 2024-05-08
feat(cli): bench pub support split file content

ysfscream:
### PR Checklist

#### Issue Number

Example: \#123

image

image

- [ ] Yes
- [ ] No

#### Specific Instructions

### DIFF
--- a/cli/src/index.ts
+++ b/cli/src/index.ts
@@ -501,6 +501,7 @@ export class Commander {
         'load the parameters from the local configuration file, which supports json and yaml format, default path is ./mqttx-cli-config.json',
       )
       .option('--file-read <PATH>', 'read the message body from the file', parseFileRead)
+      .option('--split <CHARACTER>', 'split the input message in a single file by a specified character.')
       .allowUnknownOption(false)
       .action(benchPub)
 
--- a/cli/src/types/global.d.ts
+++ b/cli/src/types/global.d.ts
@@ -114,6 +114,7 @@ declare global {
     messageInterval: number
     limit: number
     verbose: boolean
+    split: string
   }
 
   type OmitSubscribeOptions = Omit<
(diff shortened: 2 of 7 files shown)


cli/src/index.ts
```diff
@@ -501,6 +501,7 @@ export class Commander {
         'load the parameters from the local configuration file, which supports json and yaml format, default path is ./mqttx-cli-config.json',
       )
       .option('--file-read <PATH>', 'read the message body from the file', parseFileRead)
+      .option('--split <CHARACTER>', 'split the input message in a single file by a specified character.')
```
Red-Asuka, line 504: Maybe we should set `\n` as the default value.

cli/src/types/global.d.ts
```diff
@@ -114,6 +114,7 @@ declare global {
     messageInterval: number
     limit: number
     verbose: boolean
+    split: string
```
Red-Asuka, line 117: This is an optional option, which may be undefined.

ysfscream: but the limit, verbose, and others are optional, too, right? Was there no impact before?

Red-Asuka: The default value of limit is `0`, and verbose is a boolean type option. The default behavior is false, so it has no effect. However, it is best to respect the actual situation and modify `verbose` to `verbose?: boolean`.

cli/src/index.ts
ysfscream: yes, I thought so at first, but I didn't finish.

cli/src/types/global.d.ts
ysfscream: so we need to set a default value for the split option as well.

Red-Asuka: It should be similar to the save and config options. If the option is not declared, it will be undefined. If the option is declared but no value is set, it will default to a default value.

Red-Asuka (approved):

PR #1697, opened 2024-06-18
chore(web): disable the ws protocol in the online version

Red-Asuka:
### PR Checklist

#### Issue Number

Example: \#123

image

image

- [ ] Yes
- [ ] No

#### Specific Instructions


web/src/lang/connections.ts
```diff
@@ -507,4 +507,9 @@ export default {
     en: 'Clear history successfully',
     ja: '履歴を正常にクリアする',
   },
+  wsProtocolNotAllowed: {
+    zh: '出于安全考虑，我们提供的在线环境不再支持使用 ws 协议。对于需要使用 ws 的用户，建议使用 Docker 本地部署。详情请参考：{link}',
+    en: 'For security reasons, the online environment we provide no longer supports the use of the ws protocol. For users who need to use ws, it is recommended to use Docker local deployment. For details, please refer to: {link}',
+    ja: 'セキュリティ上の理由から、提供するオンライン環境ではwsプロトコルの使用をサポートしていません。wsを使用する必要があるユーザーは、Dockerのローカルデプロイメントをお勧めします。詳細については、次を参照してください：{link}',
+  },
```
ysfscream, line 514: To enhance user experience and provide additional solutions, it is recommended to update the existing message with the following suggestions 👇:

Add the suggestion to download the MQTTX desktop or CLI version from https://mqttx.app/. We recommend Docker for local deployment for users who need to use the web version.

web/src/utils/mqttUtils.ts
```diff
@@ -155,7 +155,7 @@ export const getDefaultRecord = (): ConnectionModel => {
     updateAt: time.getNowDate(),
     name: '',
     clean: true,
-    protocol: 'ws',
+    protocol: process.env.BASE_URL === '/web-client/' ? 'wss' : 'ws',
```
ysfscream, line 158: **Suggested Improvement:**
Could you consider adding an explicit environment variable or a configuration setting to distinguish between online and local deployments? This will provide a more robust solution.

Example:
1. Add an environment variable, e.g., `IS_ONLINE_ENV`.
2. Update the protocol determination logic accordingly.

```javascript
protocol: process.env.IS_ONLINE_ENV === 'true' ? 'wss' : 'ws',
```

Using an explicit flag like `IS_ONLINE_ENV` ensures that the protocol selection is clear and unambiguous, regardless of how the base URL is configured.

ysfscream (approved):

PR #1713, opened 2024-07-07
feat(cli): allow specifying the default protocol in config files

rpendleton:

Previously, config files couldn't specify the default protocol.

Config files can now specify the default protocol.

The init command has also been updated to prompt for the protocol, and it also recommends a different default port based on the selected protocol.

- [ ] Yes
- [x] No

#### Other information

Although I updated the init command to recommend / use a different default port based on the protocol, I didn't update the default port to be based on protocol anywhere else in the app, since I thought that might be too much of a breaking change.

Similarly, if you manually create a config file with a protocol other than `mqtt` but don't specify a port, it will use the default port of 1883 instead of what was recommended by the init command. We could maybe determine the default port based on the protocol in this case without it being a breaking change (since the protocol config property is new), but I worry whether that will lead to unexpected inconsistencies compared to specifying a protocol with command line flags.

Because of this, I'm not sure whether it actually makes sense for the init script to suggest a different port based on the selected protocol, but I figure it's worth at least discussing.

(And then as an aside, there might be other defaults that should be overridable using a config file, such as websocket path, ca cert, alpn, etc. I didn't have a need for any of those properties in my configuration though, so I only added the protocol for now. Alternatively, it might make sense to use a URI for some of these parameters. That would be a more involved change though.)


cli/src/configs/common.ts
```diff
@@ -13,10 +13,27 @@ const DEFAULT_CONFIG: ConfigModel = {
...
 const VALID_OUTPUT_MODES: Array<ConfigModel['output']> = ['text', 'log']
 
-export { USER_HOME_DIR, CONFIG_FILE_PATH, DEFAULT_CONFIG, VALID_OUTPUT_MODES }
+const VALID_PROTOCOLS: Array<Protocol> = ['mqtt', 'mqtts', 'ws', 'wss']
+
+const RECOMMENDED_PORTS_BY_PROTOCOL: Record<Protocol, number> = {
+  mqtt: 1883,
+  mqtts: 8883,
+  ws: 80,
+  wss: 443,
```
ysfscream, line 29: Thank you for this PR!

I recommend updating the default ports for `ws` and `wss` to `8083` and `8084`, respectively. These ports are more commonly used by popular MQTT brokers like EMQX, which might make them more intuitive for our users.

rpendleton: I wasn't completely sure about ws and wss, so that's good to know! Updated.

ysfscream:
So many thanks to you! I will merge it.

ysfscream (approved):

PR #1715, opened 2024-07-08
feat(desktop): support setup cli to windows path

ysfscream:
### PR Checklist

Windows users can only run `mqttx.exe` manually. They should find the CLI exe file first on the system.

#### Issue Number

Example: None

 Add `${downloadPath}/mqttx.exe` to the PATH to run `mqttx.exe` anywhere.

https://stackoverflow.com/questions/59560042/how-to-append-values-to-the-path-environment-variable-in-nodejs

- [ ] Yes
- [x] No

#### Specific Instructions

### DIFF
--- a/src/main/installCLI.ts
+++ b/src/main/installCLI.ts
@@ -25,8 +25,8 @@ async function checkInstalledMqttxCLI(win: BrowserWindow, isWindows: boolean): P
   }
   return new Promise((resolve) => {
     exec('mqttx --version', (error, stdout, stderr) => {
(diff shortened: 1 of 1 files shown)


ysfscream:
Waiting for real testing on the Windows system.

ysfscream:
@Red-Asuka Updating the PATH environment variable is a sensitive and dangerous operation that must be done cautiously. Therefore, we will temporarily not support using the PATH method to update the Windows CLI. I will close this PR.

Issue #1723, opened 2024-07-17
The connection name cannot be adjusted


engine100:
#### What did I do

create a connection , the name is long

#### What happened

The connection name on the left is not fully displayed and cannot be adjusted

macos  : 14.2.1 
emqx: Version: v1.10.1

ysfscream:
Thanks for your feedback. We will fix it in the next version.

ysfscream:
Updated on https://github.com/emqx/MQTTX/releases/tag/v1.11.0, thanks!

PR #1724, opened 2024-07-17
fix(desktop): fix the style issue of connection name

XL-YiBai:
### PR Checklist

image
image

#### Issue Number

\#1723

image

- [ ] Yes
- [x] No

#### Specific Instructions

### DIFF
--- a/src/views/connections/ConnectionsDetail.vue
+++ b/src/views/connections/ConnectionsDetail.vue
@@ -30,9 +30,16 @@
                 <i class="iconfont icon-show-connections"></i>
               </a>
             </el-tooltip>
-            <h2 :class="{ offline: !client.connected }">
-              <span class="title-name">{{ titleName }}</span>
-            </h2>
+            <el-tooltip
+              :effect="theme !== 'light' ? 'light' : 'dark'"
+              :content="`${titleName}`"
+              :open-delay="500"
+              placement="top"
+            >
+              <h2 :class="['title-name', !client.connected ? 'offline' : undefined]">
+                {{ titleName }}
+              </h2>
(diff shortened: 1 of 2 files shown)


Red-Asuka (approved):

src/views/connections/ConnectionsDetail.vue
```diff
@@ -30,9 +30,16 @@
...
+            <el-tooltip
+              :effect="theme !== 'light' ? 'light' : 'dark'"
+              :content="`${titleName}`"
+              :open-delay="500"
+              placement="top"
+            >
+              <h2 :class="[ { offline: !client.connected }, 'title-name' ]">
+                {{ titleName }}
+              </h2>
+            </el-tooltip>
```
ysfscream, line 42: So, it always shows this tooltip for the title, right?

But I think it's good to show it when the title is too long. Maybe you can set a limit number?

src/views/connections/ConnectionsDetail.vue
```diff
@@ -30,8 +30,19 @@
                 <i class="iconfont icon-show-connections"></i>
               </a>
             </el-tooltip>
-            <h2 :class="{ offline: !client.connected }">
-              <span class="title-name">{{ titleName }}</span>
+            <el-tooltip
+              v-if="showTitleTooltip"
```
ysfscream, line 34: The current implementation for displaying the tooltip using v-if is a bit complex. we could use the `disabled` props.

XL-YiBai: Oh, yes! I didn't notice this prop ~

ysfscream (approved):

PR #1735, opened 2024-08-19
feat(cli): add avro support

LAST7:
### PR Checklist

Support only protobuf as schema-based encoding. (cli)

#### Issue Number

None

Support avro encoding now. (cli)

- [ ] Yes
- [x] No

#### Other information

I wonder what tests should be written to test the function and stability.

### DIFF
--- a/cli/src/index.ts
+++ b/cli/src/index.ts
@@ -245,6 +245,11 @@ export class Commander {
         '-Pmn, --protobuf-message-name <NAME>',
         'the name of the protobuf message type (must exist in the .proto file)',
       )
+      .option(
+        '-Ap, --avsc-path <PATH>',
+        'the path to the .avsc file that defines the avro schema for AVRO decoding',
+        parseFileRead,
+      )
       .option('--debug', 'enable debug mode for MQTT.js', false)
       .allowUnknownOption(false)
       .action(pub)
@@ -366,6 +371,11 @@ export class Commander {
       .option(
         '-Pmn, --protobuf-message-name <NAME>',
         'the name of the protobuf message type (must exist in the .proto file)',
+      )
+      .option(
+        '-Ap, --avsc-path <PATH>',
+        'the path to the .avsc file that defines the avro schema for AVRO decoding',
+        parseFileRead,
       )
       .option('--debug', 'enable debug mode for MQTT.js', false)
       .allowUnknownOption(false)
--- a/cli/src/index.ts
+++ b/cli/src/index.ts
@@ -247,7 +247,7 @@ export class Commander {
       )
       .option(
         '-Ap, --avsc-path <PATH>',
-        'the path to the .avsc file that defines the avro schema for AVRO decoding',
+        'the path to the .avsc file that defines the avro schema for AVRO encoding',
         parseFileRead,
       )
       .option('--debug', 'enable debug mode for MQTT.js', false)
--- a/cli/src/lib/pub.ts
+++ b/cli/src/lib/pub.ts
@@ -13,6 +13,7 @@ import logWrapper, { basicLog, benchLog, Signale, signale, simulateLog, singaleC
 import { handleLoadOptions, handleSaveOptions } from '../utils/options'
 import { checkScenarioExists, checkTopicExists, parseConnectOptions, parsePublishOptions } from '../utils/parse'
 import { serializeProtobufToBuffer } from '../utils/protobuf'
+import { serializeAvroToBuffer } from '../utils/avro'
 import { loadSimulator } from '../utils/simulate'
 
 /**
@@ -25,18 +26,16 @@ import { loadSimulator } from '../utils/simulate'
  * Flow:
  *   Input Message -> [Format Conversion] -> [Protobuf Serialization] -> Output Message
  * @param {string | Buffer} message - The message to be processed.
- * @param {string} [protobufPath] - The path to the protobuf definition.
- * @param {string} [protobufMessageName] - The name of the protobuf message.
+ * @param {SchemaOptions} [schemaOptions] - Options for schema-based encoding
  * @param {FormatType} [format] - The format to convert the message to.
  * @returns {Buffer | string} - The processed message.
  */
 const processPublishMessage = (
   message: string | Buffer,
-  protobufPath?: string,
-  protobufMessageName?: string,
+  schemaOptions: SchemaOptions,
   format?: FormatType,
 ): Buffer | string => {
-  const convertMessageFormat = (msg: string | Buffer): Buffer | string => {
+  const convertMessageFormat = (msg: string | Buffer): string | Buffer => {
     if (!format) {
       return msg
     }
@@ -44,16 +43,22 @@ const processPublishMessage = (
     return convertPayload(bufferMsg, format, 'encode')
   }
 
-  const serializeProtobufMessage = (msg: string | Buffer): Buffer | string => {
-    if (protobufPath && protobufMessageName) {
-      return serializeProtobufToBuffer(msg.toString(), protobufPath, protobufMessageName)
+  const serializeWithSchema = (msg: string | Buffer): string | Buffer => {
+    switch (schemaOptions.type) {
+      case 'none':
+        return msg
+
+      case 'protobuf':
+        return serializeProtobufToBuffer(msg, schemaOptions.protobufPath, schemaOptions.protobufMessageName)
+
+      case 'avro':
+        return serializeAvroToBuffer(msg, schemaOptions.avscPath)
     }
-    return msg
   }
 
-  const pipeline = [convertMessageFormat, serializeProtobufMessage]
+  const pipeline = [convertMessageFormat, serializeWithSchema]
 
-  return pipeline.reduce((msg, transformer) => transformer(msg), message) as Buffer
+  return pipeline.reduce((msg: string | Buffer, transformer) => transformer(msg), message) as Buffer
 }
 
 const send = (
@@ -62,8 +67,7 @@ const send = (
   pubOpts: {
     topic: string
     message: string | Buffer
-    protobufPath: string | undefined
-    protobufMessageName: string | undefined
+    schemaOptions: SchemaOptions
     format: FormatType | undefined
     opts: IClientPublishOptions
   },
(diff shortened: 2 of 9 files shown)


cli/src/lib/pub.ts
```diff
@@ -25,35 +26,39 @@ import { loadSimulator } from '../utils/simulate'
  * Flow:
  *   Input Message -> [Format Conversion] -> [Protobuf Serialization] -> Output Message
```
ysfscream, line 27: Sorry, could you please update this flow comment? Although it is not from you.

cli/src/lib/pub.ts
```diff
@@ -25,35 +26,39 @@ import { loadSimulator } from '../utils/simulate'
...
+    switch (schemaOptions.type) {
+      case 'none':
+        return msg
+
+      case 'protobuf':
+        return serializeProtobufToBuffer(msg, schemaOptions.protobufPath, schemaOptions.protobufMessageName)
+
+      case 'avro':
+        return serializeAvroToBuffer(msg, schemaOptions.avscPath)
     }
```
ysfscream, line 56: I prefer using a default clause to simplify the switch statement:

```typescript
switch (schemaOptions.type) {
  case 'protobuf':
    return serializeProtobufToBuffer(msg, schemaOptions.protobufPath, schemaOptions.protobufMessageName)
  case 'avro':
    return serializeAvroToBuffer(msg, schemaOptions.avscPath)
  default:
    return msg
}
```

cli/src/lib/sub.ts
```diff
@@ -17,27 +18,45 @@ import * as Debug from 'debug'
  * Flow:
  *  payload -> [Protobuf Deserialization] -> [Format Conversion] -> Processed Message
```
ysfscream, line 19: and same here.

paylaod -> schema [protobuf, avro] -> xxx

cli/src/lib/sub.ts
```diff
@@ -17,27 +18,45 @@ import * as Debug from 'debug'
...
+          payload,
+          schemaOptions.protobufPath,
+          schemaOptions.protobufMessageName,
+          format,
+        )
+
+      case 'avro':
+        return deserializeBufferToAvro(payload, schemaOptions.avscPath, format)
+    }
+  }
```
ysfscream, line 55: and the same here. I prefer using a default clause here.

cli/src/types/global.d.ts
```diff
@@ -56,6 +56,23 @@ declare global {
...
+  interface AvroSchemaOptions {
+    type: 'avro'
+    avscPath: string
+  }
+
+  interface NoSchema {
+    type: 'none'
+  }
+
+  type SchemaOptions = ProtobufSchemaOptions | AvroSchemaOptions | NoSchema
```
ysfscream, line 74: # Code Review: Simplifying SchemaOptions

I prefer simplifying the `SchemaOptions` type and the `serializeWithSchema` function:

```typescript
interface ProtobufSchemaOptions {
  type: 'protobuf';
  protobufPath: string;
  protobufMessageName: string;
}

interface AvroSchemaOptions {
  type: 'avro';
  avscPath: string;
}

type SchemaOptions = ProtobufSchemaOptions | AvroSchemaOptions;

const processPublishMessage = (
  message: string | Buffer,
  schemaOptions?: SchemaOptions,
  format?: FormatType,
): Buffer | string => {
  // ... existing code ...

  const serializeWithSchema = (msg: string | Buffer): string | Buffer => {
    if (!schemaOptions) return msg;
    switch (schemaOptions.type) {
      case 'protobuf':
        return serializeProtobufToBuffer(msg, schemaOptions.protobufPath, schemaOptions.protobufMessageName);
      case 'avro':
        return serializeAvroToBuffer(msg, schemaOptions.avscPath);
    }
  }

  // ... rest of the function ...
}
```

Key simplifications:

1. Removed the 'none' type from `SchemaOptions`. This reduces type declarations and simplifies the type system.
2. Made `schemaOptions` optional in the function parameter (`schemaOptions?: SchemaOptions`). This allows `undefined` to be passed, replacing the need for a 'none' type.
3. In `serializeWithSchema`, we now check if `schemaOptions` is falsy (which covers both `undefined` and `null`), returning the original message if so. This replaces the functionality of the 'none' type.
4. Removed the `default` case in the `switch` statement, as all cases are now explicitly handled.

From my side, i think these changes make the code more concise and easier to understand while maintaining the existing functionality and allowing for the absence of schema options.

cli/src/index.ts
```diff
@@ -245,6 +245,11 @@ export class Commander {
         '-Pmn, --protobuf-message-name <NAME>',
         'the name of the protobuf message type (must exist in the .proto file)',
       )
+      .option(
+        '-Ap, --avsc-path <PATH>',
+        'the path to the .avsc file that defines the avro schema for AVRO decoding',
```
ysfscream, line 250: I think the schema will be used to `encoding` for the sub command, not `decoding`. am i right?

cli/src/utils/avro.ts
```diff
@@ -0,0 +1,59 @@
...
+import * as fs from 'fs'
+import logWrapper from './logWrapper'
+
+const schemaCache: { [key: string]: avro.Type } = {}
+
+const getAvroType = (schemaPath: string): avro.Type => {
+  // first search from cache
+  if (schemaCache[schemaPath]) {
+    return schemaCache[schemaPath]
+  }
```
ysfscream, line 11: It's a good idea for the performance improvement. but I'm questioning the purpose of this cache. MQTTX CLI operations are typically independent, and the memory should be released after each operation is completed.

This means the cache variable won't persist between runs. Since we're not persisting this data, the cache will be empty whenever we start MQTTX. So, it doesn't seem very useful. Are there other scenarios where this cache might be needed?

cli/src/utils/avro.ts
```diff
@@ -0,0 +1,59 @@
...
+}
+
+export const serializeAvroToBuffer = (raw: string | Buffer, avscSchemaPath: string): Buffer => {
+  const type: avro.Type = getAvroType(avscSchemaPath)
+
+  let rawMessage = raw.toString('utf-8')
+
+  try {
+    const serializedMessage = type.toBuffer(JSON.parse(rawMessage))
```
ysfscream, line 33: **Suggestion:** Consider JSON validation before Avro conversion

While errors are caught during schema conversion, separate JSON parsing could improve error debugging. Optional, as the current implementation still catches errors.

```typescript
try {
  JSON.parse(rawMessage);
} catch (err) {
  throw new Error(`Invalid JSON input: ${err.message}`);
}
```

cli/src/utils/avro.ts
```diff
@@ -0,0 +1,59 @@
...
+
+const getAvroType = (schemaPath: string): avro.Type => {
+  // first search from cache
+  if (schemaCache[schemaPath]) {
+    return schemaCache[schemaPath]
+  }
+
+  try {
+    const schemaStr = fs.readFileSync(schemaPath, 'utf-8')
+    const type = avro.Type.forSchema(JSON.parse(schemaStr))
```
ysfscream, line 15: Same here: Consider JSON validation before Avro conversion. but it's an optional.

cli/src/lib/pub.ts
```diff
@@ -25,35 +26,39 @@ import { loadSimulator } from '../utils/simulate'
...
  * @param {FormatType} [format] - The format to convert the message to.
  * @returns {Buffer | string} - The processed message.
  */
 const processPublishMessage = (
   message: string | Buffer,
-  protobufPath?: string,
-  protobufMessageName?: string,
+  schemaOptions: SchemaOptions,
   format?: FormatType,
 ): Buffer | string => {
```
Red-Asuka, line 37: I believe that `schemaOptions` should be an optional parameter, and that `schemaOptions` and `format` should be combined into a single options object.

The code could be refactored like this:

```ts
const processReceivedMessage = (
  payload: Buffer,
  options?: {
    format?: "base64" | "json" | "hex" | "cbor" | "binary"
    schema: 'protobuf' | 'avro'
    protobufPath?: string
    protobufMessageName?: string
    avscPath?: string
  },
): string | Buffer => {
```

The rationale is that these parameters are all related to processing the `payload` and are optional. By merging them into a single options object, we eliminate concerns about argument order and make the code more intuitive and easier to maintain.

cli/src/utils/avro.ts
LAST7: I noticed that if we start mqttx(cli) with a '-sub' option, it will run in the terminal continously and receive messages from the broker. In this situation, each message received would require a 'fs.fileReadSync' if there's no cache mechanism. So I think this should have some effect on improving performance.

cli/src/lib/pub.ts
LAST7: Thank you for your effort on reviewing my code. But I have a different opinion on the design of this parameter.
1. `format` and `schemaOptions` serve different purposes within the processing pipeline.(They are used in different functions)
2. merging them into one parameter would bypass the restriction set in type `SchemaOptions`, which I believe is kind of unsafe.

But I do understand your suggestion on simplifying the signature of this function, maybe we can come to a middle ground which retains safety while simplifies the signature.

cli/src/utils/avro.ts
ysfscream: Good. I forgot the sub-command was still running for the code.

cli/src/utils/parse.ts
```diff
@@ -358,6 +359,20 @@ const parsePublishOptions = (options: PublishOptions) => {
...
+      type: 'protobuf',
+      protobufPath,
+      protobufMessageName,
+    }
+  } else if (avscPath) {
+    schemaOptions = {
+      type: 'avro',
+      avscPath,
+    }
+  }
```
ysfscream, line 374: It seems like the SchemaOptions construction is written twice? Do we need to modify this part into a function? If we add new Schema types in the future, would we only need to modify the implementation within the function?

same here: https://github.com/emqx/MQTTX/pull/1735/files#diff-dd4b60885f3579b04fe2d1a24931010150551d1389e4139efa6f3252d37d2243R162-R174

ysfscream:
@LAST7 Thanks for your contribution and everything LGTM.

ysfscream (approved):

Red-Asuka (approved):

Red-Asuka:
@LAST7 Thanks for the PR!

PR #1743, opened 2024-08-28
feat(desktop): support enable ignore qos 0 messages

ysfscream:
### PR Checklist

All messages will be stored

#### Issue Number

Example: None

According to the MQTT protocol specification, it supports ignoring QoS messages and not storing them locally. When enabled, this can optimize performance for receiving and publishing large volumes of messages

image

- [ ] Yes
- [ ] No

#### Specific Instructions

### DIFF
--- a/src/views/connections/ConnectionsDetail.vue
+++ b/src/views/connections/ConnectionsDetail.vue
@@ -1332,12 +1332,8 @@ export default class ConnectionsDetail extends Vue {
   private async saveMessage(id: string, messages: MessageModel[]) {
     try {
       if (messages.length) {
-        const filteredMessages = messages.filter((msg) => !ignoreQoS0Message(msg.qos))
-        console.log('filteredMessages', filteredMessages)
-        if (filteredMessages.length) {
-          const { messageService } = useServices()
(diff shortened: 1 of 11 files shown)


src/views/connections/ConnectionsDetail.vue
```diff
@@ -1332,8 +1332,12 @@ export default class ConnectionsDetail extends Vue {
   private async saveMessage(id: string, messages: MessageModel[]) {
     try {
       if (messages.length) {
-        const { messageService } = useServices()
-        await messageService.pushMsgsToConnection(messages, id)
+        const filteredMessages = messages.filter((msg) => !ignoreQoS0Message(msg.qos))
```
Red-Asuka, line 1335: I believe it would be more efficient to place the QoS filtering after processing `nonSYSMessageSubject` and then use `filter` within RXJS's `pipe` for further filtration. This approach consolidates all data processing under RXJS, providing a unified handling method.

ysfscream: exactly!

Red-Asuka (approved):

PR #1757, opened 2024-09-10
feat(desktop): add avro support

LAST7:
### PR Checklist

- [ ] Yes
- [x] No

#### Specific Instructions

### DIFF
--- a/src/components/UseScript.vue
+++ b/src/components/UseScript.vue
@@ -123,7 +123,7 @@ export default class UseScript extends Vue {
     this.functions = functions
 
     this.schemas =
-      this.currentSchemaType == null
+      this.currentSchemaType === null
         ? []
         : schemas.filter((s) => s.name.endsWith(this.defaultSchemaTypes[this.currentSchemaType!].extension))
   }
--- a/src/components/UseScript.vue
+++ b/src/components/UseScript.vue
@@ -24,8 +24,17 @@
 
         <el-col :span="24">
           <el-form-item :label="$t('script.functionName')" prop="currentFunctionId">
-            <el-select filterable clearable size="small" v-model="currentFunctionId">
+            <el-select filterable clearable size="small" v-model="currentFunctionId" @change="handleFunctionChange">
               <el-option v-for="func in functions" :key="func.id" :value="func.id" :label="func.name"> </el-option>
+            </el-select>
+          </el-form-item>
+        </el-col>
+
+        <el-col :span="24">
+          <el-form-item :label="$t('script.schemaType')" prop="todo">
+            <el-select filterable clearable size="small" v-model="currentSchemaType" @change="handleSchemaTypeChange">
+              <el-option v-for="schema in schemaList" :key="schema.label" :value="schema.value" :label="schema.label">
+              </el-option>
             </el-select>
           </el-form-item>
         </el-col>
@@ -39,7 +48,7 @@
           </el-form-item>
         </el-col>
 
-        <el-col :span="24" v-if="currentSchemaId">
+        <el-col :span="24" v-if="currentSchemaId && currentSchemaType == 'protobuf'">
           <el-form-item :label="$t('script.protoName')" prop="currentProtoName">
             <el-input v-model.trim="currentProtoName" size="mini"></el-input>
           </el-form-item>
@@ -63,9 +72,14 @@ export default class UseScript extends Vue {
   @Prop({ default: false }) public visible!: boolean
 
   private showDialog: boolean = this.visible
+  private schemaList: SchemaList[] = [
+    { label: 'Protobuf', value: 'protobuf' },
+    { label: 'Avro', value: 'avro' },
+  ]
   private functions: ScriptModel[] = []
   private schemas: ScriptModel[] = []
   private currentFunctionId: string = ''
+  private currentSchemaType: SchemaType | null = null
   private currentSchemaId: string = ''
   private currentProtoName: string = ''
   private scriptApply: MessageType = 'all'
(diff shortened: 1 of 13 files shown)


src/components/UseScript.vue
```diff
@@ -93,24 +107,51 @@ export default class UseScript extends Vue {
...
+
   private async loadData() {
     const { scriptService } = useServices()
     const functions: ScriptModel[] | [] = (await scriptService.getAllFunction()) ?? []
     const schemas: ScriptModel[] | [] = (await scriptService.getAllSchema()) ?? []
     this.functions = functions
-    this.schemas = schemas
+
+    this.schemas =
+      this.currentSchemaType == null
```
ysfscream, line 126:
```suggestion
      this.currentSchemaType === null
```

Although `==` is fine, but `===` is a better lint for JS.

src/utils/schemaUtils.ts
```diff
@@ -0,0 +1,12 @@
+export const convertObject = (raw: string, format?: PayloadType): string => {
```
ysfscream, line 1: There is a bit of confusion. Does the function of this utility function seem to be able to use the convertPayload function? 🤔

ysfscream:
There aren't any unit test cases at the moment. It's possible to run tests now. Could you consider adding more Avro tests to enhance and improve its stability?

ysfscream:
You can paste some usage screenshots in the comments to clarify the feature review, although we can check it locally. I think there is no big problem on my side with these code changes. However, I think I need to do more local testing after your PR has been merged.

LGTM. Thank you!

src/views/script/index.vue
```diff
@@ -203,12 +214,33 @@ export default class Script extends Vue {
...
   // TODO: Because the editor does not support the parsing of the protobuf language, the editor language is temporarily set to javascript.
   private currentSchema: SchemaType = 'protobuf'
   private currentFunction: FunctionType = 'javascript'
   private readonly inputTypeList: PayloadType[] = ['JSON', 'Plaintext', 'Base64', 'Hex']
+
+  get uploadButtonLabel(): string {
+    // function tab or schema tab
+    switch (this.activeTab) {
+      case 'functionTab':
+        return this.$t('script.uploadJs') as string
```
Red-Asuka, line 231: For the old version of i18n, you can use `$tc` to return a string type, thus avoiding the use of type assertion.

src/views/connections/ConnectionsDetail.vue
```diff
@@ -1806,42 +1807,99 @@ export default class ConnectionsDetail extends Vue {
...
+            to,
+          )
         }
+
+        default:
+          return payload
       }
-      return convertedPayload
     } catch (error) {
       this.$message.error((error as Error).toString())
```
Red-Asuka, line 1901: Although it might not significantly impact the program, for error handling in try-catch, you could refer to this solution on StackOverflow: https://stackoverflow.com/a/64452744/18652097.

Red-Asuka (approved):

src/utils/schemaUtils.ts
LAST7: This is function originates from `src/utils/protobuf.ts`, and it seems to be cover only some of the formats supported in `convertPayload`.
Actually, I am also confused as it being apart from the `convertPayload` function.

It looks more similiar to the `convertPayload` in the *cli* part(`cli/src/utils/convertPayload.ts`) rather than the one in *desktop* part(`src/utils/convertPayload.ts`).

The logic of encoding/decoding message with format from these two components(cli & desktop) are quite different, take Base64 as example:

desktop:

```TypeScript
const convertBase64 = (value: string, codeType: 'encode' | 'decode'): string => {
  const convertMap: CodeType = {
    encode(str: string): string {
      return btoa(unescape(encodeURIComponent(str)))
    },
    decode(str: string): string {
      return decodeURIComponent(escape(atob(str)))
    },
  }
  return convertMap[codeType](value)
}

```

cli:

```TypeScript
const convertPayload = (payload: Buffer | string, format?: FormatType, action: Action = 'decode') => {
  const actions = {
    encode: {
      // It looks like its decoding from a base64 buffer
      base64: () => Buffer.from(payload.toString(), 'base64'),
      json: () => convertJSON(payload, 'encode'),
      ...

```

Actually, the methods inside cli's `convertPayload` look also utterly strange. Cuz to me, this `base64: () => Buffer.from(payload.toString(), 'base64')` seems to be decoding a base64 encoded message into a normal Buffer, which is strange as it's not performing any 'encode' action.

I think we should have an online discussion about these format conversion.

LAST7:
> You can paste some usage screenshots in the comments to clarify the feature review, although we can check it locally. I think there is no big problem on my side with these code changes. However, I think I need to do more local testing after your PR has been merged.
> 
> LGTM. Thank you!

I will post a video. Just a moment

src/views/connections/ConnectionsDetail.vue
LAST7: It's a good idea to avoid returning undefined with such approach. But it seems that I would have to change the return type of the function to `any` or `unknown` to allow this object-like thing to be returned, which makes it nasty when validating the return variable of this function in the upper process.
Can you provide more detail on what return type should be stated and how to validate this `{ message: ... }`? Thank you so much!

Red-Asuka: Specific to this example, I don't think you need to make any modifications, `toString` can be applied to the vast majority of data types.

LAST7: Ah, you mean the `error as Error` thing.
I do agree that it's better to state it in the `catch (error instanceof Error)`, but in most cases the project do `(error as Error).message` or `(error as Error).toString()`, so I think following the same pattern would be a better choice. 
After all, there's no significant difference between these two code style. :)

Red-Asuka: Yes, for this kind of simple error handling scenario, using type assertion is indeed more convenient. I just want to remind that this type of error handling may sometimes have hidden risks.

The try...catch block can catch errors of any type, not just objects that inherit from Error. For example, a piece of code may throw a string, a number, or even an object, which do not necessarily inherit from the Error class.

By using type assertion like error as Error, you may lose some actual information about the error. If the caught exception is not an instance of Error, this type assertion will result in missing important information in your subsequent handling, or even runtime errors.

TypeScript is designed to avoid unnecessary type assertions because excessive use of type assertions can make the code more fragile and harder to maintain. By leveraging TypeScript's type inference and type checking features as much as possible, such issues can be reduced.

Here I can give a relatively complex example, such as error handling for API calls, where we can encapsulate a generic handling method like this:

```ts
/**
 * Handles fetch errors.
 *
 * @param {unknown} error - The caught error object.
 * @param {() => boolean} [cb] - Optional callback function. If the callback exists and returns true, the default error handling will not be executed.
 */
export function handleFetchError(error: unknown, cb?: () => boolean) {
  if (error instanceof FetchError) {
    if (cb && cb())
      return

    const err = error as ResError
    const { statusText } = err.response || {}
    const { detail } = err?.response?._data || {}
    if (detail ?? statusText)
      ElMessage.error(detail ?? statusText)
    else
      console.error('Fetch error: ', error)
  }
  else {
    console.error('Unknown error: ', error)
  }
}

LAST7:
The size of input/output textboxes in Script -> Schema tab are expanded:

Shot-2024-09-14-112224

src/views/connections/ConnectionsDetail.vue
LAST7: I see. Now I understand the hidden risk of using type assertion here, thank you for the explanation!
We could improve all the try-catch blocks in the project later for better error handling and safety.

ysfscream (approved):

PR #1758, opened 2024-09-11
chore(config): add experimental option in tsconfig to fit modern lsp

LAST7:
modern lsp like volar would complain about variables used in template without this option.

### PR Checklist

image

image

**error in line 23 is already addressed in another pr.**

- [ ] Yes
- [x] No

#### Other information

Haven't tried it on VSCode, but since I'm using the same lsp as VSCode popular plugin uses, it should behave the same.

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -33,6 +33,7 @@
   },
   "vueCompilerOptions": {
     "target": 2,
+    "experimentalUseElementAccessInTemplate": true,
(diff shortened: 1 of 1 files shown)


ysfscream:
@LAST7, thank you so much for your contribution! LGTM. However, I noticed that this configuration seems to be a deprecated item. But it's fine for the moment.

image

PR #1766, opened 2024-09-24
feat(client): allow client id empty

ysfscream:
### PR Checklist

It is impossible to simulate MQTT 5 when the Client ID is empty. In this case, the MQTT Broker can automatically allocate a Client ID.

https://www.emqx.com/zh/blog/mqtt5-connect-properties

#### Issue Number

Example: https://askemq.com/t/topic/7181

Support clearing the client ID, and do not limit the client ID to a required field.

image

- [ ] Yes
- [x] No

#### Specific Instructions

### DIFF
--- a/src/utils/mqttUtils.ts
+++ b/src/utils/mqttUtils.ts
@@ -53,6 +53,7 @@ export const getClientOptions = (record: ConnectionModel): IClientOptions => {
     clientIdWithTime,
   } = record
   const protocolVersion = mqttVersionDict[mqttVersion as '3.1' | '3.1.1' | '5.0']
+
   const options: IClientOptions = {
     clientId,
     keepalive,
--- a/src/views/connections/ConnectionForm.vue
+++ b/src/views/connections/ConnectionForm.vue
@@ -72,35 +72,6 @@
               </el-tooltip>
             </el-col>
             <el-col :span="22">
-              <el-form-item label-width="93px" label="Client ID" prop="clientId">
-                <el-input size="mini" v-model="record.clientId"></el-input>
-              </el-form-item>
-            </el-col>
-            <el-col :span="1">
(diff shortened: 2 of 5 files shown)


Red-Asuka:
@ysfscream The requirements for Client ID vary across different versions of MQTT. It might be better to dynamically validate and provide guidance on Client ID usage based on the selected MQTT version.

> In MQTT, the handling of the Client ID differs across versions:
> 
> MQTT 3.1 and 3.1.1:
> 
> Clients are required to provide a Client ID, which is a unique identifier.
> In MQTT 3.1, if no Client ID is provided, the connection will be rejected by the broker.
> In MQTT 3.1.1, clients are allowed to send an empty Client ID, but only if the "Clean Session" (or "Clean Start") flag is set. If not, the connection will be rejected. When using an empty Client ID with Clean Session, the broker will generate a temporary Client ID for the session, but the connection state won't be persisted.
> MQTT 5.0:
> 
> MQTT 5.0 introduces more flexibility regarding Client ID management. Clients can send an empty Client ID, and in this case, the broker will automatically generate a Client ID for the client.
> This enhancement simplifies the connection management for clients that don't need to manually specify an ID.
> In summary:
> 
> In MQTT 3.1 and 3.1.1, a Client ID is generally required (with 3.1.1 allowing it to be empty under specific conditions).
> In MQTT 5.0, clients can omit the Client ID, and the broker will automatically assign one.

ysfscream:
@Red-Asuka I have considered this, but for EMQX, all MQTT Versions are supported. After testing, I found that if the MQTT Broker is not supported, or when the Clean Session is False, the Client ID required error message is also very clear, so I don't want to put too many restrictions on the Client ID being empty, because the default is the value of the Client ID, and only special needs will clear the client ID.

Issue #1810, opened 2024-11-14
[Bug] MQTTX CLI  -- mqttx conn hangs after connection


oising:
#### What did I do

Whether passing via explicit args, or using saved json config, `mqttx conn` hangs after connecting. Is this meant to hang, or open a shell? What is the point of this functionality? If it's meant to test connection, it should close once connected successfully. Very confusing.

#### What happened

mqttx hangs.

![image](https://github.com/user-attachments/assets/6125380f-93c0-423f-94fa-e08df50009ee)

#### Expected

Absolutely no idea.

#### Environment

windows 11 arm64, SL7, node 22

#### More detail

More detail include `log` information or screenshot

ysfscream:
The `mqttx conn` behavior is working as intended. It's designed to establish and maintain an MQTT connection - MQTT is built around sessions and persistent connections. The "hanging" state indicates a successful active connection.

You can:
- Use Ctrl+C to disconnect when done
- Use `pub` for publishing messages
- Use `sub` for subscribing
- Try `bench` or `simulate` for advanced scenarios

If you just need to verify connectivity, you can Ctrl+C after a successful connection.

oising:
You could make it more explicit my emitting "Press ctrl+c to exit"

ysfscream:
> You could make it more explicit my emitting "Press ctrl+c to exit"

Good suggestions, thank you.

ysfscream:
Try https://github.com/emqx/MQTTX/releases/tag/v1.11.1 Thanks.

PR #1821, opened 2024-12-02
feat(cli): add exit tips for conn

ysfscream:
### PR Checklist

When users successfully establish a connection using the `mqttx conn` command, the program maintains the connection state (appearing to be "hanging"). Still, there is no prompt for users to exit. This causes users to need clarification about the program's behavior.

- Users are unclear that the expected behavior of the `mqttx conn` command is to maintain the connection
- Users don't know how to exit the connection state properly
- Lack of user guidance leads to poor operational experience

#### Issue Number

Example: #1810 

```shell
❯ mqttx conn
✔ Connected
- Press Ctrl+C to disconnect and exit
```

- [ ] Yes
- [x] No

#### Specific Instructions

### DIFF
--- a/cli/src/utils/exitInfo.ts
+++ b/cli/src/utils/exitInfo.ts
@@ -1,0 +1,5 @@
+import chalk from 'chalk'
+
+export const triggerExitInfo = () => {
+  console.log(chalk.dim('- Press Ctrl+C to disconnect and exit'))
+}
--- a/cli/src/lib/conn.ts
+++ b/cli/src/lib/conn.ts
@@ -4,6 +4,7 @@ import { parseConnectOptions } from '../utils/parse'
 import delay from '../utils/delay'
 import { handleSaveOptions, handleLoadOptions } from '../utils/options'
 import * as Debug from 'debug'
+import { triggerExitInfo } from '../utils/exitInfo'
 
 const conn = (options: ConnectOptions) => {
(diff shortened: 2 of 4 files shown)


cli/src/utils/exitInfo.ts
```diff
@@ -0,0 +1,5 @@
+import chalk from 'chalk'
+
+export const triggerExitInfo = () => {
+  console.log(chalk.dim('- Press Ctrl+C to disconnect and exit'))
```
Red-Asuka, line 4: You can use different prompts according to the operating system.

```ts
// Import the 'os' module using ES Module syntax
import os from 'os';

// Get the current platform
let platform = os.platform();

// Declare a variable for the exit message
let exitMessage;

// Check the platform type and set the appropriate exit message
if (platform is 'win32') {
    exitMessage = 'Press Ctrl+C to exit';
} else if (platform is 'darwin') {
    exitMessage = 'Press ⌘+C to exit';
} else {
    exitMessage = 'Press Ctrl+C to exit';
}
```

ysfscream: But every OS uses `Ctrl + C`. There is no ⌘...

Red-Asuka: Looks like my brain temporarily switched to 'another operating system' mode. Just a minor glitch! 😂

Red-Asuka (approved):

PR #1826, opened 2024-12-09
feat(links): integrate app version and update link management

Red-Asuka:
This pull request includes several changes to handle the application version dynamically and improve the localization of links based on the user's locale. The most important changes include importing the version from `package.json`, defining a global `__APP_VERSION__` variable, and updating the `useLinks` composable to provide localized links.

Handling application version dynamically:

* [`apps/desktop/electron.vite.config.ts`](diffhunk://#diff-10a3bbe3228b3f2afec6048e8a29ec7d9cbf0bd8e70c93d8c0505b88fd237263R12-R13): Imported the `version` from `package.json` and defined the `__APP_VERSION__` variable in the Vite configuration. [[1]](diffhunk://#diff-10a3bbe3228b3f2afec6048e8a29ec7d9cbf0bd8e70c93d8c0505b88fd237263R12-R13) [[2]](diffhunk://#diff-10a3bbe3228b3f2afec6048e8a29ec7d9cbf0bd8e70c93d8c0505b88fd237263R22-R24)
* [`apps/web/vite.config.ts`](diffhunk://#diff-eea049695b0188b525a44f51269fe670485ed3e355f8cede185cafedec24d786R14-R20): Imported the `version` from `package.json` and defined the `__APP_VERSION__` variable in the Vite configuration.
* [`packages/types/base.ts`](diffhunk://#diff-21ba0b59740447448c94b099ca99ccef4732a5ea0ebc23d19c1c4ef625becd6bR1-R6): Declared a global `Window` interface with the `__APP_VERSION__` property.

Localization improvements:

* [`packages/ui/src/composables/useLinks.ts`](diffhunk://#diff-46f5b893a3cd2e065d03a3b610a80adf9756e78a0a85b50cc395cfb2dee14b0bL2-R53): Updated the `useLinks` composable to provide localized links based on the user's locale.
* [`packages/ui/src/composables/__test__/useLinks.test.ts`](diffhunk://#diff-586e5f0db359f5cc9cebb7e9e2a9529234cf2919ff9a6c76ad0d9a4839e8e84fR1-R136): Added tests to verify the correct links are returned based on the locale.

Other changes:

* [`packages/ui/src/components/HelpView.vue`](diffhunk://#diff-bab65da4ec4815b2bdde9400f0f2a882646900b7fa7ae864979f85ed706b5c8aL2-R9): Updated the help view to display the `help` links.
* [`packages/ui/src/components/common/LeftMenu.vue`](diffhunk://#diff-4c301db7f5e560026e2b30a663dc978c8a8f0027384fefa5f2b6a4028f36c492L10-R10): Updated the left menu to use the `leftBarLogo` link. [[1]](diffhunk://#diff-4c301db7f5e560026e2b30a663dc978c8a8f0027384fefa5f2b6a4028f36c492L10-R10) [[2]](diffhunk://#diff-4c301db7f5e560026e2b30a663dc978c8a8f0027384fefa5f2b6a4028f36c492L49-R49)
* [`scripts/gen-version.js`](diffhunk://#diff-d3e4148d13325c3ac611eff497a27c1d422e031042e0712b886beb06b3eadf52L1-L6): Removed the script that generated a `version.ts` file.

### DIFF
--- a/apps/desktop/electron.vite.config.ts
+++ b/apps/desktop/electron.vite.config.ts
@@ -9,6 +9,8 @@ import { ElementPlusResolver } from 'unplugin-vue-components/resolvers'
 import Components from 'unplugin-vue-components/vite'
 import VueRouter from 'unplugin-vue-router/vite'
 
+import { version } from './package.json'
+
 export default defineConfig({
   main: {
     plugins: [externalizeDepsPlugin()],
@@ -17,6 +19,9 @@ export default defineConfig({
     plugins: [externalizeDepsPlugin()],
   },
   renderer: {
+    define: {
+      __APP_VERSION__: JSON.stringify(version),
+    },
     resolve: {
       alias: {
         '@database': resolve('src/database'),
--- a/packages/ui/src/composables/__test__/useLinks.test.ts
+++ b/packages/ui/src/composables/__test__/useLinks.test.ts
@@ -1,13 +1,138 @@
-import { describe } from 'vitest'
+import type { Lang } from 'mqttx'
+import { mount } from '@vue/test-utils'
+import { describe, expect, it } from 'vitest'
+import { defineComponent, h, nextTick } from 'vue'
+import { createI18n } from 'vue-i18n'
+import useLinks from '../useLinks'
 
 describe('useLinks', () => {
-  // it('should initialize linksMap correctly', () => {
-  //   const { MQTTXSite } = useLinks()
+  const i18n = createI18n({
+    locale: 'en',
+    messages: {
+      en: {},
+      zh: {},
+      ja: {},
+      hu: {},
+      tr: {},
+    },
+  })
 
-  //   const expectedLinksMap = ref({
-  //     homepage: 'https://mqttx.app',
-  //   })
+  const setupI18n = (locale: Lang) => {
+    i18n.global.locale = locale
+    return nextTick()
+  }
 
-  //   expect(MQTTXSite.value).toEqual(expectedLinksMap.value)
-  // })
+  const createTestComponent = () => {
+    return defineComponent({
+      setup() {
+        return useLinks()
+      },
+      render() {
+        return h('div')
+      },
+    })
+  }
+
+  it('should return correct MQTTXSite based on locale', async () => {
+    await setupI18n('zh')
+    const wrapper = mount(createTestComponent(), { global: { plugins: [i18n] } })
+    expect(wrapper.vm.MQTTXSite).toBe('https://mqttx.app/zh')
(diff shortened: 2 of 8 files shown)


apps/desktop/electron.vite.config.ts
```diff
@@ -17,6 +19,9 @@ export default defineConfig({
     plugins: [externalizeDepsPlugin()],
   },
   renderer: {
+    define: {
+      __APP_VERSION__: JSON.stringify(version),
```
ysfscream, line 23: Why JSON stringify version?

Red-Asuka: The reason for using `JSON.stringify` in the define option is to ensure that the version is correctly treated as a string constant. Although UserConfig.define accepts Record<string, any>, the values are directly replaced in the code. If version is not a string, it could lead to a syntax error in JavaScript.

For example:

Without JSON.stringify, if version is 1.0.0, the replacement might result in:

```ts
const version = 1.0.0; // This causes a syntax error
```

With JSON.stringify, the replacement will be:

```ts
const version = "1.0.0"; // This is valid JavaScript code
```

Thus, using JSON.stringify ensures that version is correctly converted to a string, avoiding potential syntax errors.

ysfscream: Is `.toString` not working for that?

Red-Asuka: `toString` is not suitable, but you can write it as `` __APP_VERSION__: `"${version}"` ``. However, I still recommend using `JSON.stringify` because the value will be directly replaced in the code. JSON.stringify can handle more data types, avoid escaping issues, and ensure consistency and security.

Red-Asuka: https://vite.dev/config/#using-environment-variables-in-config

packages/ui/src/composables/__test__/useLinks.test.ts
ysfscream, line 1: Adding a link 404 check, I think, would make it even more complete and help validate doc links.

```js
const checkLink = async (url) => {
  if (url === '') return true
  try {
    const { status } = await axios.head(url, { timeout: 30000 })
    return status >= 200 && status < 400
  } catch (error) {
    console.error(`error link url: ${url}`)
    return false
  }
}
```

Then use Promise.all() to check all links in parallel:

```js
const results = await Promise.all(links.map((link) => checkLink(link)))
results.forEach((isValid) => {
  expect(isValid).toBe(true)
})
```

ysfscream (approved):

PR #1852, opened 2025-01-03
feat(desktop): save and restore window state

ni00:

Won't save and restore window position and size.

#### Issue Number

 #1849 

Save and restore window position and size.

- [ ] Yes
- [x] No

#### Other information

reference:  https://github.com/electron/electron/issues/526

### DIFF
--- a/src/background.ts
+++ b/src/background.ts
@@ -1,6 +1,6 @@
 import 'reflect-metadata' // Required by TypoORM.
 ;('use strict')
-import { app, protocol, BrowserWindow, ipcMain, shell, Menu } from 'electron'
+import { app, protocol, BrowserWindow, ipcMain, shell, Menu, screen } from 'electron'
 import { createProtocol } from 'vue-cli-plugin-electron-builder/lib'
 import installExtension, { VUEJS_DEVTOOLS } from 'electron-devtools-installer'
 import { quitAndRenameLogger } from './utils/logger'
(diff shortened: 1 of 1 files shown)


src/background.ts
```diff
@@ -185,6 +185,22 @@ async function createWindow() {
...
+    // Check if the window is within the bounds of the screen
+    if (
+      savedBounds.x >= screenArea.x &&
+      savedBounds.x <= screenArea.x + screenArea.width &&
+      savedBounds.y >= screenArea.y &&
+      savedBounds.y <= screenArea.y + screenArea.height
+    ) {
+      win.setBounds(savedBounds)
+    }
+  }
```
ysfscream, line 202: Could you please write to the main folder independently of the background file?

ysfscream:
So many thanks for implementing the window state persistence feature!

The implementation looks thoughtful and handles the multi-monitor scenario well. Here are some thoughts that might be helpful:

1. We might want to consider adding maximized state handling?

```typescript
// Example approach:
electronStore.set('windowState', {
  bounds: win.getBounds(),
  isMaximized: win.isMaximized()
})
```

2. Would it make sense to add some error handling? Something like:
```typescript
try {
  win.setBounds(savedBounds)
} catch (error) {
  console.error('Failed to restore window bounds:', error)
  // Handle fallback
}
```

4. As the app grows, we might want to consider moving the window state logic into its own module, but that's not urgent for now

ni00:
图片

I'm considering removing the height and width fields from SettingEntity

Reasons:

1. These two fields currently only provide default values and cannot be modified, which is inconsistent with other configurable properties in SettingEntity that can be adjusted in the settings page

2. Window size, position, and other state properties are better suited for internal application state management, and don't need to be included in user-configurable settings

3. Following practices from other Electron applications, window states are typically managed using lightweight storage solutions like electron-store, rather than being written to the main database that requires backup

4. These window state properties are not noticeable to users and don't need to be synchronized in backups

Red-Asuka:
@ni00 Thank you very much for your contribution to the project! I have optimized the existing code and resolved the building error. I completely agree with your views on storing window width and height data. In version v1.x, we will no longer use the width and height from the database table. These have already been removed in version v2.x. For more details on the changes, please see: https://github.com/emqx/MQTTX/pull/1868

Red-Asuka (approved):

ysfscream (approved):

ysfscream:
Thanks to all @ni00 @Red-Asuka

PR #1868, opened 2025-02-10
improve(desktop): improve window state management

Red-Asuka:
### PR Checklist

- Window width and height are recorded in a database table.
- Only window size can be restored, but position and fullscreen status cannot be recovered.

- Use store to record window dimensions.
- Add boundary check to ensure the window is within a visible area.
- Record fullscreen status.

- [ ] Yes
- [x] No

#### Specific Instructions

### DIFF
--- a/apps/desktop/src/main/config/windowState.ts
+++ b/apps/desktop/src/main/config/windowState.ts
@@ -23,27 +23,27 @@ export const defaultWindowSize = {
 export function restoreWindowState(win: BrowserWindow): void {
   try {
     const savedState = store.get('windowState') as WindowState | undefined
-    if (savedState && typeof savedState === 'object') {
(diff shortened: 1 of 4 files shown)


apps/desktop/src/main/config/windowState.ts
```diff
@@ -0,0 +1,65 @@
...
+  isMaximized: boolean
+  isFullScreen: boolean
+}
+
+// Default window size
+export const defaultWindowSize = {
+  width: 1025,
+  height: 749,
+  minHeight: 650,
+  minWidth: 997,
```
ysfscream, line 19: Could you consideru consider moving defaultWindowSize to a centralized constants file to improve maintainability and consistency across t?e project? This will make it easier to update and reuse window size configurations.

ysfscream (approved):

PR #1897, opened 2025-03-17
fix(desktop): fix sqlite disconnect warning issue

ysfscream:
### PR Checklist

When closing the application, an unhandled promise rejection occurs with error: `SQLITE_MISUSE: Database handle is closed`. This happens because the application attempts to close the database connection multiple times during the shutdown process, causing errors when trying to close an already closed connection.

Added safeguards to prevent multiple database connection close attempts:
1. Added a flag to track if connections have already been destroyed
2. Enhanced the connection destroy function with proper error handling
3. Modified app shutdown flow to avoid duplicate cleanup calls
4. Improved error handling in database connection management

This prevents the unhandled promise rejection and ensures a clean application shutdown.

- [ ] Yes
- [x] No

#### Specific Instructions

No special instructions needed. This fix addresses the database connection closure error without changing any application functionality.

#### Other information

The error occurred because both the window's `close` event handler and the app's `window-all-closed` event handler were attempting to close database connections, leading to the same connection being closed twice.

### DIFF
--- a/src/background.ts
+++ b/src/background.ts
@@ -20,7 +20,6 @@ import ORMConfig from './database/database.config'
 import version from '@/version'
 import { initialize } from '@electron/remote/main'
 import { initMCPHandlers, cleanupMCPConnections } from './main/ai/mcp/MCPManager'
-// import installExtension, { VUEJS_DEVTOOLS } from 'electron-devtools-installer'
 
 declare const __static: string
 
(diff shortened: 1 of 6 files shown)


src/background.ts
```diff
@@ -246,14 +255,17 @@ async function createWindow() {
...
-    try {
-      await installExtension(VUEJS_DEVTOOLS)
-    } catch (e) {
-      console.error('Vue Devtools failed to install:', e)
-    }
-  }
+  // if (isDevelopment && !process.env.IS_TEST) {
+  //   // Install Vue Devtools - modified to handle errors better
+  //   try {
+  //     await installExtension(VUEJS_DEVTOOLS, {
```
Red-Asuka, line 261: Why does this need to be commented out?

ysfscream: image

Because I get this error when launching app, and I don't have a good solution for it at the moment, no matter how I try, I can't use Vue Dev Tools. For now, the chances of me using it are very low, so I have commented it out. Perhaps it will be more useful in the 2.0 support for Vue 3? What do you think?

Red-Asuka: You might find this link helpful: https://github.com/electron-react-boilerplate/electron-react-boilerplate/issues/2577#issuecomment-1690769017

Red-Asuka: Vue2 Devtools ID: iaajmlceplecbljialhhkmedjlpdblhp

ysfscream: OK, cool, it works now, thanks.

Red-Asuka (approved):

Issue #1900, opened 2025-03-18
[Bug] Does not filter when incoming messages


kaangrbz:
#### What did I do

I open the search panel with CTRL + F and write something to message input

#### What happened

It filters when i click enter but not when incoming message came

#### Expected

Check topic and message inputs if not empty it should

#### Environment

- OS: OS version, Windows 10 Pro 22H2 19045.5608
- MQTTX version, v1.9.8

#### More detail

More detail include `log` information or screenshot

![Image](https://github.com/user-attachments/assets/95ebf7ed-8893-49df-af93-f85c6b3fabbd)

PR #1939, opened 2025-06-01
fix(connections): apply active search filters on incoming messages

muzaffarmhd:
### PR Checklist

Incoming messages bypass active search filters unless Enter is pressed again. 
#### Issue Number

#1900 

- Incoming messages are filtered based on active topic/payload filters
- `batchUpdateMsgs()` updated to apply `searchMessage()` before rendering

- [ ] Yes  
- [x] No

#### Specific Instructions

- Test filtering under high message rates  
- Verify async changes for race/UI issues  
- Confirm backward compatibility and message buffering

#### Other information

- None

### DIFF
--- a/src/views/connections/ConnectionsDetail.vue
+++ b/src/views/connections/ConnectionsDetail.vue
@@ -1452,15 +1452,22 @@ export default class ConnectionsDetail extends Vue {
   }
 
   // Push messages to the list
-  private batchUpdateMsgs(incomingMsgs: MessageModel[]) {
+  private async batchUpdateMsgs(incomingMsgs: MessageModel[]) {
     let _messages = _.cloneDeep(this.recordMsgs.list)
-    incomingMsgs.forEach((msg: MessageModel) => {
+    for (const msg of incomingMsgs) {
       const isActiveTopicMessages = matchTopicMethod(this.activeTopic, msg.topic)
       const isActiveMsgType = this.isMessageTypeActive(msg)
       if (isActiveMsgType && (!this.activeTopic || isActiveTopicMessages)) {
-        _messages.push(msg)
+        if (this.searchParams.topic !== '' || this.searchParams.payload !== '') {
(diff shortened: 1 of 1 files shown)


Red-Asuka:
@muzaffarmhd Thank you for your contribution! I noticed an optimization opportunity: since `matchMultipleSearch` is currently implemented as an async function, it causes async contagion effects that lead to some unnecessary modifications in the related code (such as using operators like `concatMap`).

After reviewing the code, I found that the `matchMultipleSearch` function internally performs purely synchronous operations (just simple array filtering and string matching) and can be completely converted to a synchronous function. This would maintain its synchronous nature while reducing code complexity and avoiding unnecessary async handling.

I suggest refactoring this function to a synchronous implementation, which would simplify the entire call chain and improve code readability and maintainability. Here's a reference implementation:

```typescript
type MatchMultipleSearch = (data: any[], params: SearchParams) => any[] | null

const matchMultipleSearch: MatchMultipleSearch = (data, params) => {
  const paramsKeys = Object.keys(params)
  try {
    const filterData = data.filter(($) => {
      return paramsKeys.every((oneKey) => {
        if ($[oneKey]) {
          const key: string = $[oneKey].toLowerCase().replace(/\s+/g, '')
          const value: string = params[oneKey]
            .toLocaleLowerCase()
            .replace(/\s+/g, '')
            .replace(/[~#^$@%&!+*]/gi, (val) => `\\${val}`)
          return key.match(value)
        } else {
          return null
        }
      })
    })
    return filterData
  } catch (error) {
    throw error
  }
}
```

muzaffarmhd:
Thanks for your review, well as you mention, it's just filtering and string matching, and the promise overhead isn't needed, for this. I'll make the necessary changes and squash the next commit into this one.

ysfscream:
@muzaffarmhd Thank you for your contribution.

ysfscream (approved):

PR #1970, opened 2025-08-19
Feat/json tree viewer

muzaffarmhd:
### PR Checklist

This Pull Request adds new features to the Viewer. Specifically follows the addition of Diff View with JSON Viewer.
Functionality wise the following features are added:
JSON Tree View: 
a. It allows user to view payloads in expandable and collapsible form using `vue-json-tree-view` as the dependency to parse and view JSON as a tree. Additionally, there is a `copy` option added which will copy the JSON payload to the system clipboard. If the payload is Non-JSON, it is displayed as a single collapsible node. 
Please check the corresponding UI:

image

Additionally, there is also a search feature implemented to search in the payload.

image

b. Visualization: We use ECharts to visualize the JSON as a tree, the depth level can be set, and on right click of node, the subtree will be copied as JSON. It also supports dark theme. 
image
c. Navigation: The payload history navigation is controlled through Arrow keys, Left and Right. 

#### Key Summary of utility function changes:
1. New utilities added:
- jsonTreeConverter: Converts a JSON structure to EChartsNode, the corresponding unit tests are added as well.
2. Modified utils:
- jsonUtils: new utility function added to stringify the ECharts subtree to JSON to avail copy functionality. Unit tests for this new utility are covered as well.

- [ ] Yes
- [ x] No

### DIFF
--- a/src/widgets/JSONViewer.vue
+++ b/src/widgets/JSONViewer.vue
@@ -2,7 +2,7 @@
   <div class="json-tree-view">
     <div class="json-header">
       <div class="header-content">
-        <h3>Message JSON Viewer</h3>
+        <h3>{{ $t('viewer.messageJsonViewer') }}</h3>
       </div>
       <div class="message-info" v-if="currentMessage">
         <div class="info-item">
@@ -38,6 +38,16 @@
                 @input="onSearchInput"
                 class="search-input"
               />
+              <el-button
+                type="primary"
+                size="small"
+                icon="el-icon-view"
+                @click="openVisualTreeModal"
+                class="visualize-btn"
+                title="Visualize JSON Tree"
+              >
+                {{ $t('viewer.visualize') }}
+              </el-button>
             </div>
             <div class="navigation-controls">
               <button
@@ -70,7 +80,7 @@
                 placement="bottom"
                 :effect="theme !== 'light' ? 'light' : 'dark'"
                 :open-delay="500"
-                :content="copiedJson ? 'Copied!' : 'Copy JSON'"
+                :content="copiedJson ? $t('common.copySuccess') : $t('common.copy')"
               >
                 <a
                   href="javascript:;"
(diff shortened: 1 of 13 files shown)


ysfscream:
> Just realized, need to test and handle Big Int issues in JSON, which is common in IoT development. Should use the existing JSON Parse method instead of JS's JSON API

Can add some edge cases/boundary conditions to the test cases for testing

src/widgets/JSONViewer.vue
```diff
@@ -0,0 +1,619 @@
...
+  private copiedJson = false
+  private searchTimeout: any = null
+  private showVisualTreeModal = false
+  private defaultExpandLevel = 1
+
+  get currentTheme(): string {
+    return this.theme || 'light'
+  }
+
+  get treeOptions(): any {
```
ysfscream, line 205: It looks like there's a fixed format here, the type shouldn't be `any`?

src/widgets/JSONViewer.vue
```diff
@@ -0,0 +1,619 @@
...
+  @Prop({ type: Boolean, default: false })
+  private hasMore!: boolean
+
+  @Getter('currentTheme')
+  private theme!: string
+
+  private searchQuery = ''
+  private currentIndex = 0
+  private copiedJson = false
+  private searchTimeout: any = null
```
ysfscream, line 197:
```suggestion
  private searchTimeout: number | null = null
```

ysfscream:
Overall I think there are no major issues, need to consider more edge cases, if there are performance issues should also consider them in advance, and there seem to be too many any types in the code

muzaffarmhd:
> > Just realized, need to test and handle Big Int issues in JSON, which is common in IoT development. Should use the existing JSON Parse method instead of JS's JSON API
> 
> Can add some edge cases/boundary conditions to the test cases for testing

I've added a utility to convert bigint instances to plain objects, if you think that is okay, i've added some test cases for it as well.

src/widgets/JSONViewer.vue
muzaffarmhd: Yes, my bad. I have fixed it. Thank you.

ysfscream:
> I've added a utility to convert bigint instances to plain objects, if you think that is okay, i've added some test cases for it as well.

It would be better if there were some test cases.

PR #1980, opened 2025-09-15
fix(KeyValueEditor.vue): fix a User Property display bug in Topics Tree

ShadowPaint-SP:
**What is the current behavior?**

When navigating in the Topics Tree from one leaf with User Properties to another one with User Properties they don't get updated. (in the topic-info-card)

**What is the new behavior?**

Now when switching from one leaf to another the User Properties get updated as disabled is set to true in this view.

**Does this PR introduce a breaking change?**

- [ ] Yes
- [x] No

### DIFF
--- a/src/components/KeyValueEditor.vue
+++ b/src/components/KeyValueEditor.vue
@@ -59,7 +59,7 @@ export default class KeyValueEditor extends Vue {
     val: ClientPropertiesModel['userProperties'],
     oldVal: ClientPropertiesModel['userProperties'],
   ) {
-    if (oldVal === undefined && val) {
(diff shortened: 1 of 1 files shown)


ysfscream:
Thanks for identifying this issue! 👍 

The fix works but using `disabled` property to control updates isn't semantically clear and might cause side effects in other scenarios.

A cleaner approach would be to directly compare the values:
```javascript
if (oldVal !== val) {
  this.processObjToArry()
}
```

This handles all value changes correctly without relying on the disabled state.

Would you like to update the code with this approach?

ysfscream (approved):
LGTM. Thanks.

PR #2002, opened 2025-11-28
ci(win): add portable version for windows

zfb132:
### PR Checklist

The Windows build configuration only outputs NSIS installers (`.exe`). It does not generate a standard portable `.zip` archive containing the unpacked application.

Added 'zip' to the `win.target` configuration in `vue.config.js`. Now, the build process will generate a portable `.zip` file for Windows (both x64 and ia32 architectures) alongside the existing installers. This provides a "green," strictly portable option without the self-extracting overhead of the portable .exe. The newly added files should be named like the following:  
```bash
MQTTX-${version}-${arch}-win.zip
```

- [ ] Yes
- [x] No

### DIFF
--- a/vue.config.js
+++ b/vue.config.js
@@ -104,9 +104,10 @@ module.exports = {
           allowToChangeInstallationDirectory: true,
           createDesktopShortcut: true,
           createStartMenuShortcut: true,
+          artifactName: 'MQTTX-Setup-${version}-${arch}.${ext}',
         },
         win: {
-          artifactName: 'MQTTX-Setup-${version}-${arch}.${ext}',
+          artifactName: 'MQTTX-${version}-${arch}-win.${ext}',
           icon: './public/icons/app.ico',
           target: [
             {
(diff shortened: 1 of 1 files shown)


ysfscream (commented):
Hi @zfb132,

Sorry for the late response. Thank you for your contribution!

A few questions before we proceed:

1. **Testing**: Have you tested that the generated zip file works correctly (i.e., the portable version runs without issues)?

2. **Artifact naming concern**: I noticed that you changed `win.artifactName` from `'MQTTX-Setup-${version}-${arch}.${ext}'` to `'MQTTX-${version}-${arch}-win.${ext}'`. Will this affect the final build output names for the existing NSIS installer?

   If it does change the installer's filename, we'd need to sync this with our website download links, which adds extra work. I'd prefer to keep the existing naming unchanged if possible. 

   Could you check if we can add the zip target without modifying the original artifact name? If it doesn't significantly impact your use case, that would be ideal.

Thanks!

PR #2015, opened 2025-12-27
fix: resolve duplicate data, optimize csv/excel export, and fix progress bar

### LINKED ISSUE #2006: [Bug] Various bugs with Export Data
Thanks for MQTTX, it's a great tool. 
However I've found that the "Export Data" function has various bugs or design flaws:

1. In JSON, YAML, or XML, each message has all of its data twice - it is at the top level of each object, and then repeated within a "properties" sub-object.

2. CSV format is not useful. A useful format would have one message per line. Instead, all messages are crammed into one line. They are all just dumped into one field in JSON format. I supect the have the same duplication problem because you are using the same JSON object. I also suspect the same is true of Excel format but I haven't been able to check that. Neither the CSV nor the Excel file of my 150,000 messages will load into Mac Numbers - it doesn't even give an error, just silently fails...

3. The Export progress bar stays on 0% for ages and then jumps straight to 100%.

An example of the duplication:
```
      {
        "id": "message_[GUID]",
        "createAt": "2025-12-05 15:48:19:837",
        "out": false,
        "payload": "This is the payload. It might be quite large. Repeating it (and all the other data) unnecessarily is probably not a good thing, as it makes the export data file twice as big as it should be.",
        "qos": 1,
        "retain": false,
        "topic": "some/topic",
        "meta": "{\"msgType\":\"Plaintext\"}",
        "payloadFormatIndicator": null,
        "messageExpiryInterval": null,
        "topicAlias": null,
        "responseTopic": null,
        "correlationData": null,
        "userProperties": null,
        "subscriptionIdentifier": null,
        "contentType": null,
        "connectionId": "[GUID]",
        "properties": {
          "id": "message_[GUID]",
          "createAt": "2025-12-05 15:48:19:837",
          "out": false,
          "payload": "This is the payload. It might be quite large. Repeating it (and all the other data) unnecessarily is probably not a good thing, as it makes the export data file twice as big as it should be.",
          "qos": 1,
          "retain": false,
          "topic": "some/topic",
          "meta": "{\"msgType\":\"Plaintext\"}",
          "payloadFormatIndicator": null,
          "messageExpiryInterval": null,
          "topicAlias": null,
          "responseTopic": null,
          "correlationData": null,
          "subscriptionIdentifier": null,
          "contentType": null,
          "connectionId": "[GUID]"
        }
      }
```

The 'properties' sub-object seems to be nothing but a duplicate of the top-level fields, so is unnecessary and could be dropped.

Design of the CSV and Excel exports needs to be re-thought to make them useful. One row per message, and clean text fields with no embedded JSON. Per-connection information is not really required if we're just trying to get the message data out.

It would be even better if we can choose which fields to export - for example I only want "topic" and "message". These are not exports that will be re-imported into MQTTX, they are exports to extract information from MQTTX's history for use elsewhere. I think perhaps you've been focused solely on using Export-Import to persist MQTTX's full internal state, but Export has many other uses than that. You could have one format that is specifically for persisting state, and then offer additional Export formats (without Import counterparts) which are for extracting information. 

Many thanks
Ben

#### Environment

- MacOS Tahoe 26.1
- MQTTX 1.12.1

ni00:
Closes #2006

### 1. Duplicate data in JSON/YAML/XML exports

Every exported message contained all fields twice — once at the top level and again inside a `properties` sub-object. Root cause: `entityToModel` used `...entity` spread into both the top-level object and `properties`, causing full duplication.

Fix: destructure MQTT5 property fields explicitly, spread only `...rest` at the top level, and build `properties` from the extracted fields only.

### 2. Broken CSV/Excel format

- All messages were crammed onto a single line rather than one row per message
- Message data was dumped as raw JSON within a single field

Fix: flatten each message into individual rows with clean columns (`client_id`, `connection_name`, `topic`, `payload`, `qos`, `retain`, `direction`, `createAt`, plus MQTT5 properties). CSV writes one row per message with proper headers; Excel creates one sheet per connection with the same flat structure.

### 3. Inaccurate progress bar

Progress was tracked per-connection (`processedConnections / totalConnections`). With 1-2 connections but 150k+ messages, the bar stayed at 0% then jumped to 100%.

Fix: count total messages before export, then update progress after each batch (~1000 messages), giving smooth incremental progress across all 5 export formats.

## Changed Files

- `src/database/services/ConnectionService.ts` — fix `entityToModel` to avoid duplicating entity fields into `properties`
- `src/database/services/MessageService.ts` — fix `entityToModel`/`modelToEntity` destructuring; add `countByConnectionId()`
- `src/main/streamExportData.ts` — rewrite CSV/Excel export to one-row-per-message; switch progress tracking to message-level granularity

## Self-tested

All cases below have been verified locally:

- [x] Export JSON — no duplicate `properties` nesting
- [x] Export YAML — clean flat structure
- [x] Export XML — no duplicate fields
- [x] Export CSV — one message per row, proper column headers, opens correctly in spreadsheet apps
- [x] Export Excel — one message per row, one sheet per connection
- [x] Export with large dataset (10k+ messages) — progress bar updates smoothly
- [x] Export with single connection — progress doesn't jump 0% → 100%

Would appreciate a second pair of eyes on these scenarios. Thanks!

### DIFF
--- a/src/database/services/ConnectionService.ts
+++ b/src/database/services/ConnectionService.ts
@@ -1,5 +1,6 @@
 import { Service } from 'typedi'
 import moment from 'moment'
+import MessageEntity from '@/database/models/MessageEntity'
 import _ from 'lodash'
 import { InjectRepository } from 'typeorm-typedi-extensions'
 import ConnectionEntity from '@/database/models/ConnectionEntity'
@@ -16,6 +17,8 @@ const electronStore = new Store()
 export const MoreThanDate = (date: string | Date) => MoreThan(DateUtils.mixedDateToUtcDatetimeString(date))
 export const LessThanDate = (date: string | Date) => LessThan(DateUtils.mixedDateToUtcDatetimeString(date))
 
+
+
 @Service()
 export default class ConnectionService {
   constructor(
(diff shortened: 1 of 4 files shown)


ysfscream (commented):
Thanks.

ysfscream (approved):
Thanks.

PR #2051, opened 2026-05-23
feat: add support for GZIP decompression

T3rm1:

Compressed content is shown as is.

In the subscription dialog there is now a checkbox "Decompress with GZIP" which decompresses all received messages with GZIP first.

image

- [ ] Yes
- [x] No


ysfscream:
Thanks for the contribution @T3rm1 — compressed payloads showing as raw bytes is a real pain point, and per-subscription granularity makes sense.

Before this can merge, could you address a few things?

1. **Model it as an encoding, not a boolean.** `processReceivedMessage` already has a documented pipeline: Raw → Schema/Type decode → Function script. GZIP is a byte-level encoding that belongs at the front of that pipeline. A `decompressGzip` checkbox does not scale (zstd/Brotli/Deflate next?). Please use an enum field like `contentEncoding: "none" | "gzip"` so the next encoding is a small addition, not a new checkbox.

2. **Decompression failure should mark the message.** A failed gunzip currently only writes to the log and the compressed bytes continue down the pipeline, so the user still sees garbage with no hint. Please use the existing `updateMetaError` mechanism (same as JSON/CBOR/MsgPack parse errors) so the message carries a visible error.

3. **Define multi-subscription matching semantics.** `subscriptions.find(matchTopic)` returns the *first* matching subscription, so with overlapping filters (e.g. `sensors/#` vs `sensors/gzip/+`) whether a message gets decompressed depends on subscription creation order. Please decompress when *any* matching subscription has the encoding set (`.some()`), or use the MQTT 5 Subscription Identifier when available for an exact match.

PR #2053, opened 2026-06-19
feat(cli): Add recevied_at timestamp for received messages

iarenaza:
When subscribing to topics to receive messages, several message details are shown (like topic, qos, size), in addition to the message payload.

But the date and time at which the message was received is not shown. Which makes mqttx CLI less useful as an unattended message recorder. Specially if the recorded messages are going to be used to diagnose unexpected or wrong communications, and correlate those communications with logs or other kind of diagnostics data from other sources.

Those correlations need, most of the time, a shared time line. The fact that mqttx CLI does not provide those timestamps makes those correlations impossible.

The current version does not display the date and time at which each message, for the subscribed topic, was received.

Now a `received_at` field is displayed, with the date and time (in ISO-8601 format, using the UTC timezone) at which a each message was received.

- [ ] Yes
- [X] No

#### Specific Instructions

No new instructions are needed. The only difference is that now `mqttx sub` also displays the `received_at` field.

### DIFF
--- a/cli/src/lib/sub.ts
+++ b/cli/src/lib/sub.ts
@@ -176,6 +176,7 @@ const sub = (options: SubscribeOptions) => {
 
     options.verbose && msgData.push({ label: 'mqtt-packet', value: packet })
 
+    msgData.push({ label: 'received_at', value: new Date().toISOString() })
     msgData.push({ label: 'topic', value: topic })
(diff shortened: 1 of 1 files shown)


ysfscream (approved):

ysfscream:
Thanks for the contribution, @iarenaza! 🙏

